PR #6, opened 2017-02-14
IDR deployment dynamic openstack inventory

manics:
Add the dynamic openstack inventory script used for internal deployments. This was previously in a private repo.

Changes from the previous private script:
- rebased on top of the latest https://github.com/ansible/ansible/blob/c53db300b887939f972226ecd2225e03164dcf7d/contrib/inventory/openstack.py
- Adds an env-var `OS_PROXY_SSH_ARGS` to set custom ssh options on the ProxyCommand gateway. For example this can be used to provision and connect to a completely new IDR system without receiving SSH key-changed errors during the deployment process by setting `OS_PROXY_SSH_ARGS='-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'`. This is a security risk and should only be enabled for the initial deployment where you know the server key will be new.
#


joshmoore:
I'm not going to be able to test this remotely due to firewall issues. Generally, :+1: for the migration. I assume the change of `inventory` to `inventories` implies that you expect this directory to be used differently?

manics:
Yes, `inventories` is going to be for example inventories only, you will not be able to use it directly with ansible-playbook.

PR #9, opened 2017-02-21
Restore 

aleksandra-tarkowska:
This Playbook allows to restore prod db from rsync

### DIFF
--- a/ansible/group_vars/all.yml
+++ b/ansible/group_vars/all.yml
@@ -1,0 +1,2 @@
+# rsync
+rsync_host: "rsync://idr-rsync.openmicroscopy.org/"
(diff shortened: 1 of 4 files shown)


manics:
can you:
- rebase this
- the added playbook isn't run routinely as part of our deployments, so how about putting it in a subdirectory?
- why do you need global variables `group_vars/all`?

aleksandra-tarkowska:
> - the added playbook isn't run routinely as part of our deployments, so how about putting it in a subdirectory?

any particular name for subdir?

> - why do you need global variables group_vars/all?

vars are scoped to host how I would make rsync shared across db and omero hosts?

manics:
> any particular name for subdir?

Something like maintenance, contrib, external, utils, .... depending on what other playbooks you're thinking of.

> how I would make rsync_host shared across db and omero hosts

I think I'd just duplicate it in the playbook since it's all self-contained.

joshmoore:
Closing for the moment.

PR #14, opened 2017-03-16
idr034 fixes

joshmoore:
* do more rsync bootstrapping
 * fix travis' molecule build

### DIFF
--- a/ansible/idr-docker.yml
+++ b/ansible/idr-docker.yml
@@ -190,29 +190,6 @@
                 USERNAME = "{{ idr_jupyter_omero_user }}"
                 PASSWORD = "{{ idr_jupyter_omero_pass | quote }}"
 
-# - name: download mineotaur data
-#   become: yes
-#   synchronize:
-#     src: "rsync://{{ TBD_omero_host }}/mineotaur/"
-#     dest: "/tmp/mineotaur/"
-
-# - name: change mineotaur to 1000 uid
-#   become: yes
-#   file:
-#     path: "/tmp/mineotaur"
-#     owner: 1000
-
-# - name: enable mineotaur container
-#   docker_container:
-#     name: mineotaur
-#     image: imagedata/mineotaur
-#     entrypoint: java
-#     volumes:
-#       - /tmp/mineotaur:/mineotaur
-#     ports:
-#       - "8080:8080"
-#     command: ["-Xmx8g", "-Djava.security.egd=file:/dev/./urandom", "-jar", "/home/build/src/target/Mineotaur-1.0.1.jar", "-start", "/mineotaur"]
-
   vars:
   - idr_jupyter_docker_swarm: "{{ docker_swarm_enabled }}"
 
--- a/ansible/idr-docker.yml
+++ b/ansible/idr-docker.yml
@@ -190,6 +190,29 @@
                 USERNAME = "{{ idr_jupyter_omero_user }}"
                 PASSWORD = "{{ idr_jupyter_omero_pass | quote }}"
 
+# - name: download mineotaur data
+#   become: yes
+#   synchronize:
+#     src: "rsync://{{ TBD_omero_host }}/mineotaur/"
+#     dest: "/tmp/mineotaur/"
+
+# - name: change mineotaur to 1000 uid
+#   become: yes
+#   file:
+#     path: "/tmp/mineotaur"
+#     owner: 1000
+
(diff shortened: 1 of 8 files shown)


ansible/idr-docker.yml
```diff
@@ -190,6 +190,29 @@
...
+#   docker_container:
+#     name: mineotaur
+#     image: imagedata/mineotaur
+#     entrypoint: java
+#     volumes:
+#       - /tmp/mineotaur:/mineotaur
+#     ports:
+#       - "8080:8080"
+#     command: ["-Xmx8g", "-Djava.security.egd=file:/dev/./urandom", "-jar", "/home/build/src/target/Mineotaur-1.0.1.jar", "-start", "/mineotaur"]
+
```
manics: Before you take this further, would you mind moving mineotaur to a separate playbook that's run after `idr-docker.yml` (or maybe included, depending on how things progress)? If it turns out the analysis docker swarm is multipurpose (jupyter and mineotaur) we should start to decouple things. This should also make it easier if we run multiple docker swarms for different purposes.

ansible/idr-proxy.yml
```diff
@@ -49,14 +49,22 @@
         }}:8080
     when: "{{ groups[idr_environment | default('idr') + '-a-dockermanager-hosts'] is defined }}"
 
-  - name: Configure rsync port in selinux
+  - name: Install packages for selinux
```
manics: Can you use https://github.com/openmicroscopy/ansible-role-selinux-utils#selinux-utils instead and make any se* tasks conditional on `selinux_enabled`.

ansible/idr-proxy.yml
```diff
@@ -49,14 +49,22 @@
...
-      ports: 873
-      proto: tcp
-      setype: http_port_t
+    yum:
+      name: "policycoreutils-python"
       state: present
 
+  - name: Configure rsync port in selinux
+    become: yes
+    command: semanage port -m -t http_port_t -p tcp 873
```
manics: This won't be idempotent. I think we should investigate why `seport` isn't working (since it could cause us problems in future with other ports), and if that doesn't work consider a conditional.

ansible/molecule-playbook.yml
```diff
@@ -46,3 +46,7 @@
     yum:
       pkg: net-tools
       state: present
+  - name: Install libselinux-python
```
manics: Remove (see previous selinux comment)

ansible/idr-proxy.yml
joshmoore: I'll roll back, but opinions on why it's not working welcome.

joshmoore:
Changes pushed; starting tests locally as well.

ansible/idr-proxy.yml
manics: Which version of Ansible? Once all the other changes are in if it still doesn't work can you add a trello card?

joshmoore:
```
~/opt/management_tools/idr/deployment/ansible $ ~/opt/management_tools/v/bin/pip freeze | grep -E "(ansible|molecule)"
ansible==2.1.4.0
ansible-lint==3.4.8
molecule==1.20.0
```

ansible/molecule-playbook.yml
```diff
@@ -2,6 +2,8 @@
 
 # Check the networking is setup as expected
 - hosts: all
+  roles:
+  - role: openmicroscopy.selinux-utils
```
manics: You shouldn't need this as it'll be installed in the standard playbook that you're testing.

joshmoore:
@manics: green.

ansible/idr-docker.yml
manics: Can you delete this since it's now in idr-mineotaur.yml

ansible/idr-mineotaur.yml
```diff
@@ -0,0 +1,27 @@
...
+  - name: change mineotaur to 1000 uid
+    become: yes
+    file:
+      path: "/tmp/mineotaur"
+      owner: 1000
+
+  - name: enable mineotaur container
+    docker_container:
+      name: mineotaur
+      image: imagedata/mineotaur
```
manics: You've added a `TBD_omero_host` above, so how about a `# TODO: TBD_version` as a reminder to tag the docker image in future?

manics:
Merging based on slack chat with @joshmoore

PR #15, opened 2017-03-24
Rewrite playbooks and vars to work with new omero-* roles (IDR-0.3.5)

manics:
This mostly involves renaming variables or changing their type (e.g. list/dict). In addition I've removed `IDR-OMERO-52-omero.j2` and put those variables into `group_vars/omero-hosts`. Many of these variables are of json type, and the new omero roles allow autoconversion of YAML lists/dicts into json config strings. This should be easier to maintain.

--depends-on (may require a tag)
- [x] https://github.com/openmicroscopy/ansible-role-omero-server/pull/8
- [x] https://github.com/openmicroscopy/ansible-role-omero-web
- [x] https://github.com/openmicroscopy/ansible-role-omero-common
- [x] https://github.com/openmicroscopy/ansible-role-omego
- [x] https://github.com/openmicroscopy/ansible-role-ice/pull/1
- [x] https://github.com/openmicroscopy/ansible-role-network-cloud-interfaces/pull/2
- [x] https://github.com/openmicroscopy/ansible-role-omero-web-apps/pull/1

Other than that there should be no difference from the previous version. OMERO.server and OMERO.web are still on the same host, future PRs may involve changing this.


joshmoore:
Is this what was deployed as test35?

joshmoore:
After a "yes" in slack: looking at the server, the results of these changes seems quite sane. There are a few questions in terms of naming, etc. to go through, but the basic goal of a tidied-up installation seems to be achieved.

joshmoore:
All boxes are checked. This should be good for a requirements bump.

manics:
Bumped, waiting for travis.

joshmoore:
Merging with travis green. Will handle final version bumps in gh-16

PR #18, opened 2017-04-17
rsync: update script to match new deployment

joshmoore:
`bin/omero` no longer looks to be on the user's
PATH. This uses the absolute paths.

### DIFF
--- a/ansible/files/idr-rsync-backup.sh
+++ b/ansible/files/idr-rsync-backup.sh
@@ -3,14 +3,14 @@
 set -e
 set -u
 
-HOST=$(su omero -lc "/opt/omero/server/OMERO.server/bin/omero config get omero.db.host")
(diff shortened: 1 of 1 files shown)


manics:
```
$ sudo bash -x ./idr-rsync-backup.sh
+ set -e
+ set -u
++ su omero -lc '/opt/omero/server/OMERO.server/bin/omero config get omero.db.host'
+ HOST=192.168.1.7
++ su omero -lc '/opt/omero/server/OMERO.server/bin/omero config get omero.db.pass'
+ PASS=idr
++ date +%Y-%m-%d
+ DATE=2017-04-18
+ test -e 2017-04-18
+ pg_dump -h 192.168.1.7 -U omero idr -j 8 -Fd -f 2017-04-18 --exclude-table-data password --exclude-table-data eventlog
Password:
```
Set `PGPASSWORD` instead?

joshmoore:
Current script is very broken. Merging and if need be we can improve for the next round.

PR #21, opened 2017-04-26
Fix uod nfs mount

manics:
The NFS mount should be `/uod/idr`

In the future the `idr-*-nfs.yml` playbooks should be refactored into a single playbook, with variables in group-vars.


joshmoore:
:+1:

PR #23, opened 2017-05-03
Set omero.web.wsgi_workers: 25 (IDR-0.3.7)

manics:
See https://trello.com/c/s22IQBKY/26-increase-gunicorn-workers

OMERO.web and OMERO.server are currently on the same VM. In the event this change to the number of wsgi-workers causes problems revert it.


joshmoore:
:+1: for testing in prod37.

joshmoore:
```
[jamoore@prod37-omero web]$ OMERO.web/bin/omero config get | grep work
omero.web.wsgi_workers=25
```

aleksandra-tarkowska:
Do we know roughly how omero server works are calculated?

joshmoore:
I do not

PR #26, opened 2017-05-05
all Cell lines should be preset by default

aleksandra-tarkowska:
List all cell lines when openening category.


joshmoore:
:+1: deployed on test37

PR #29, opened 2017-06-06
Prometheus monitoring

manics:
Opening this to give a preview of what I've done so far.

### DIFF
--- a/ansible/management-prometheus.yml
+++ b/ansible/management-prometheus.yml
@@ -115,4 +115,4 @@
     - https://idr.openmicroscopy.org/about/
 
     prometheus_rsync_banner_targets:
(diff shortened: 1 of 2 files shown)


joshmoore:
Couple of quick comments from reviewing the transferred repos:
 * 0.1.0 for all three?
 * some still mention `idr-demo`
 * ...thought I had another one. Maybe after lunch...

manics:
@joshmoore How would you feel about merging this as is (so not yet run by default).

joshmoore:
Removed use of `idr-rsync`. Otherwise, uses public tags so :+1: for merging even before this is integrated into the standard deploy.

manics:
--rebased-to #32

PR #33, opened 2017-06-23
OMERO-readonly and analysis scaling

manics:
- Uses a single network instead of separate public and analysis networks
- One OMERO read-write and two OMERO read-only servers
- Three node kubernetes cluster (one master, two workers) replaces the former single docker host
- JupyterHub and Mineotaur are deployed on Kubernetes


ansible/group_vars/dockermanager-hosts.yml
```diff
@@ -1,6 +1,6 @@
 docker_use_ipv4_nic_mtu: True
 
-#idr_jupyter_hub_image: "imagedata/jupyterhub-githubauth:0.3.0"
-idr_jupyter_notebook_image: "imagedata/jupyter-docker:0.4.0"
+idr_jupyter_hub_image: "manics/jupyterhub-githubauth:merge"
```
joshmoore: Ready to bump?

ansible/group_vars/dockerworker-hosts.yml
```diff
@@ -0,0 +1,4 @@
+docker_use_ipv4_nic_mtu: True
+
+idr_jupyter_hub_image: "manics/jupyterhub-githubauth:merge"
```
joshmoore: Ready to bump?

ansible/group_vars/omero-hosts.yml
```diff
@@ -1,10 +1,20 @@
 ######################################################################
+# Variables common to OMERO read-write and read-only servers
+######################################################################
+
+
+######################################################################
 # Shared variables
-idr_omero_release: "0.4.0"
+idr_omero_release: "OMERO-build"
```
joshmoore: Working to get this for you now for a final bump

joshmoore:
@manics : CYCP https://github.com/joshmoore/deployment/commit/47cc53444c7fd35d6962d1c4845cbb711fda759a

manics:
- [ ] --depends-on https://github.com/openmicroscopy/ansible-role-haproxy/pull/2
- [ ] --depends-on https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/4

manics:
Coming back to your previous points:

> What are your plans for the kubernetes role?

Following our previous discussion on roles I'm treating this as "in development" for now. I tried a few existing Kubernetes ansible roles without success, though I've since discovered yet another one (https://github.com/kubernetes-incubator/kubespray) which may work. Depending on how it goes `roles/kubernetes` will either be made into a standalone role, or removed.

> Is there some content here for a read-only-omero role?

Too early to tell. This should become clearer once we've specified which of the following are required:
- `omero.cluster.read_only`: `true``false`
- OMERO data directories: read-write, read-only
- OMERO database: read-write, read-only

> The kubernetes-spec directory feels like it's in the wrong place.

Agreed, though at the moment some of the spec files are templated and therefore need to be generated by Ansible. If possible I'd like to avoid this.

manics:
This should be complete. The travis failure is due to Jekyll https://github.com/jekyll/jekyll/issues/6221

ansible/group_vars/proxy-hosts.yml
```diff
@@ -12,16 +12,25 @@ nginx_proxy_worker_processes: 4
...
 ######################################################################
 # OMERO.web proxy
 
+nginx_proxy_log_format: main_timed_cache_upstream
+
 nginx_proxy_websockets_enable: True
 
+nginx_proxy_upstream_servers:
+- name: omeroreadonly
+  balance: ip_hash
```
joshmoore: `balance: ip_hash`

I hadn't realized this was the strategy. So assuming we want to downtime an RO, we'd need to update the nginx config first (and likely haproxy?) and _then_ shutdown the OMERO, right?

ansible/idr-haproxy.yml
```diff
@@ -0,0 +1,62 @@
...
+    set_fact:
+      omero_omeroreadonly_hosts: >-
+        {{
+          groups[idr_environment | default('idr') + '-omeroreadonly-hosts'] |
+          map('extract', hostvars,
+            ['ansible_' + (idr_net_iface | default('eth0')), 'ipv4', 'address']) | sort
+        }}
+
+  # https://stackoverflow.com/a/35608380
+  # TODO: Find a nicer way to do this instead of a set_fact loop
```
joshmoore: What's the general plan for handling these TODOs?

ansible/group_vars/proxy-hosts.yml
manics: `ip_hash` is the only way in the open-source version of Nginx to ensure a client with a session is sent to the same backend, but I think Nginx should detect a backend is down and automatically "recalculate" the hash.

Haproxy and the commercial version of Nginx Plus both have proper support for HTTP session pinning.

ansible/idr-haproxy.yml
manics: Hope someone comes up with an answer? I don't have one. I could change it to `RFE` or some other tag you can think of?

joshmoore: Guess I mean _your_ plan. Are these all of the same severity? Does anything need carding?

manics: The majority of the TODOs are already know by us, and related to the incomplete read-only work. I don't think more cards are needed.

joshmoore: :+1:

PR #34, opened 2017-07-20
IDR 0.4.3 merge branch

manics:
I've annotated commits with `git notes` (though GitHub doesn't support them) so they can be split into separate PRs later.

--exclude


manics:
This should all be merged as separate PRs now

PR #37, opened 2017-08-04
Add grafana dashboards (IDR-0.4.2)

manics:
Adds Grafana to the management VM, and auto-configures some dashboards.

### DIFF
--- a/ansible/grafana-dashboards/idr-per-server.json
+++ b/ansible/grafana-dashboards/idr-per-server.json
@@ -1,0 +1,195 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "Prometheus",
+      "description": "",
+      "type": "datasource",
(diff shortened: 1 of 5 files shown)


joshmoore:
Nice! How close are we to being able to make these available without SSH-port forwarding? It would significantly increase their usability.

manics:
Proxying would be easy, it's the security aspects I'm worried about. At the moment we know the only entry to everything apart from OMERO.web and jupyter is via a SSH tunnel.

joshmoore:
And http auth over https?

manics:
I've added an issue https://github.com/IDR/deployment/issues/49

joshmoore:
Fair enough, but for me this falls under the "JRS usability" header (similar to the shared directory on jupyter) so we should target it as soon as possible.

(Less of a priority but a bigger impact would be having this/these as `openmicroscopy.*` roles, but one thing at a time :smile:)

PR #38, opened 2017-08-04
OMERO.web performance/errors (IDR-0.4.2)

manics:
Changes made as a response to failures in release testing: https://trello.com/c/isuehtZ6/78-release-testing-failures
- Uses a separate Redis server for each OMERO.web instance. This shouldn't be necessary but was done to rule out as many potential issues as possible, it may be reverted in future.
- Enables the little known never tested `omero.web.public.cache.enabled` option. See https://trello.com/c/Di7IgStr/80-omerowebpubliccacheenabled and linked conversations


joshmoore:
I think the multiple redises is going to bite us quite soon. (i.e. as soon as we try to use a micro-service). Merging this as the state of 0.4.2 but likely testing a single redis should come as soon as 0.4.3

PR #39, opened 2017-08-04
Send slack messages to per-deployment channels (IDR-0.4.2)

manics:
This changes the destination channels:
- `idr-notify` to `idr-notify-XXX` (munin notifications)
- `idr-logs` to `idr-logs-XXX` (OMERO server/web error logs)
- `idr-deployment` to `idr-logs-XXX` (docker notifications)


joshmoore:
Do the channels get auto-created?

manics:
Unfortunately not. It might be possible https://api.slack.com/methods/channels.join but I'd have to check whether it requires extra privileges.

joshmoore:
WFM

PR #40, opened 2017-08-04
Use a shared OMERO idr-metadata directory (IDR-0.4.2)

manics:
The idr-metadata directory is intrinsically linked to the OMERO ManagedRepository, so the read-only servers should use the read-write server's copy.


joshmoore:
NB: with the recent discussion of cleaning up symlinks, it may be easiest to find all `/uod/idr/metadata` etc. symlinks under `/data/OMERO/ManagedRepository` and point them to `/data/idr-metadata`.

cc: @eleanorwilliams

PR #36, opened 2017-08-04
More prometheus monitoring (IDR-0.4.2)

manics:
Adds more Prometheus monitoring, some of which was used to investigate performance problems in the openmicroscopy `metadata53`branch:
- Installs a custom version of `django_prometheus`
  - Changes were made to `django_prometheus` to work with multi-process wsgi/gunicorn (the alternative was code changes to OMERO.web)
  - Subsequently someone has opened an upstream PR with the same functionality: https://github.com/korfuri/django-prometheus/pull/46 so we can revert to upstream when it's released
- Re-enables the jmx-exporter, using a custom build
  - The problems we had with the jmx-exporter should be fixed by https://github.com/prometheus/jmx_exporter/pull/162 so I made a devel build to test this: https://github.com/openmicroscopy/ansible-role-prometheus-jmx/pull/1
  - Subsequently there was an upstream release https://github.com/prometheus/jmx_exporter/releases/tag/parent-0.10 so we can revert to upstream in the next deployment after testing
- Adds a small python util for monitoring the number of omero sessions: https://gist.github.com/manics/238a63cfd9c2fb8a450252d79e609296

Requirements:
- [x] --depends-on https://github.com/openmicroscopy/ansible-role-prometheus/pull/4
- [x] --depends-on https://github.com/openmicroscopy/ansible-role-prometheus-jmx/pull/1

PR #41, opened 2017-08-04
Front-end proxy (IDR-0.4.2)

manics:
- Increases the timeout in omeroreadwrite's OMERO.web nginx.conf
  - In future this should be replaced by a template that takes advantage of https://github.com/openmicroscopy/openmicroscopy/pull/5387

Note Nginx proxy changes are also present in https://github.com/IDR/deployment/pull/36


joshmoore:
Other than a reminder that we probably need to `grep TODO` at some point, :+1:

PR #43, opened 2017-08-04
Update delete script with omero cluster and flat net (IDR-0.4.2)

manics:
We've re-architected the IDR with OMERO.readonly and a distributed VAE.


joshmoore:
Makes sense to me.

PR #44, opened 2017-08-04
Pin Kubernetes/kubeadm version (IDR-0.4.2)

manics:
An error was introduced in kubeadm 1.7.1: https://github.com/kubernetes/kubeadm/issues/347
This pins the Kubernetes version to 1.7.0 until it's fixed in an upstream release.

### DIFF
--- a/ansible/group_vars/dockermanager-hosts.yml
+++ b/ansible/group_vars/dockermanager-hosts.yml
@@ -1,5 +1,7 @@
 docker_use_ipv4_nic_mtu: True
 
+kubernetes_kube_version: 1.7.0
+
 idr_jupyter_hub_image: "imagedata/jupyterhub-githubauth:0.4.0"
 idr_jupyter_notebook_image: "imagedata/jupyter-docker:develop"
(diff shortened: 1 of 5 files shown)


joshmoore:
Shall we use an issue on this repo to remember to come back to 347 ?

manics:
I'm currently looking at other Kubernetes deployment tools for high-availability, I see this as a stopgap. Ideally we'd use an official ansible role but I couldn't get them to work.

manics:
👍  though in hindsight I think we probably should pin an important requirement like Kubernetes anyway, in which case this is just one more dependency to track.

joshmoore:
Yeah, definitely need to automate the "check for version updates".

cc: @sbesson @jburel 

So you want to leave this as 1.7.0 then or go for the newer?

manics:
`1.7.0` since that's the last time this playbook was tested.

joshmoore:
Agreed. Going to be quite interesting culling all repos to find all specifications of versions...

PR #57, opened 2017-10-19
Add experimental-omero-fpbioimage.yml playbook (IDR-0.4.4)

manics:
This is not run by default, and will require more work if we want this in production.

### DIFF
--- a/ansible/experimental-omero-fpbioimage.yml
+++ b/ansible/experimental-omero-fpbioimage.yml
@@ -1,0 +1,30 @@
+# Experimental: fpbioimage
+
+- hosts: "{{ idr_environment | default('idr') }}-omero-hosts"
+
+  roles:
+  # Needed for handlers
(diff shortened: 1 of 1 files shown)


jburel:
That will be useful to have it not only for IDR.
Maybe something to consider as a good example post training sessions i.e. maybe a role per app 
and an example playbook for the various apps.
Could you indicate the work you are thinking of for production ready so it can be turned into a detailed exercise?

manics:
Performance is very poor due to the serial loading of planes on initialisation (20+ minutes in the case of a lightsheet example). In the case of the IDR I can improve that by adding Nginx caching.

jburel:
plane loading is a problem that's true

sbesson:
Discussed with @jburel, integrating here for now although this might be better extracted as a separate role/playbook outside IDR.

PR #58, opened 2017-10-19
More grafana dashboards. Enable anonymous access (IDR-0.4.4)

manics:
Anonymous access doesn't require a login, but it still allows read-access to all data. This must be reviewed if it is to be publicly proxied.

### DIFF
--- a/ansible/grafana-dashboards/idr-postgresql.json
+++ b/ansible/grafana-dashboards/idr-postgresql.json
@@ -1,0 +1,297 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "prometheus",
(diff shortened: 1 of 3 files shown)


joshmoore:
This _PR_ needs reviewing (before merging) or _Grafana_ must be reviewed before opening it up?

manics:
I meant we need to review all information that's exposed _if_ we make direct anonymous access possible. At the moment only non-interactive images are public https://github.com/IDR/deployment/pull/66

joshmoore:
I'm still not following. Can you outline what you'd like tested/reviewed here?

manics:
Nothing really, the comment was a warning not to make this directly accessible to the public

PR #55, opened 2017-10-19
Docker/kubernetes host deployment (IDR-0.4.4)

manics:
Removes notebooks from the docker hosts- this is now handled inside the `imagedata/jupyter-docker` notebook image.

PR #59, opened 2017-10-19
Major refactoring of Jupyterhub (IDR-0.4.4)

manics:
This is a significant change to Jupyterhub. Notebooks are now managed entirely within the notebook image instead of on the server. All stoarge (`scratch` and `shared`) are now managed dynamically within Kubernetes.

--depends-on https://github.com/IDR/deployment/pull/55


ansible/kubernetes-spec/jupyterhub/deployment.yml
```diff
@@ -58,39 +58,35 @@ spec:
...
     metadata:
       labels:
         name: jupyterhub-pod
-        confighash: "{{ kubernetes_jupyterhub_config_hash }}"
     spec:
       serviceAccount: jupyterhub-spawner
       containers:
       - name: jupyterhub-container
-        image: manics/jupyterhub-githubauth:merge
+        image: imagedata/jupyterhub-githubauth:latest
```
joshmoore: Needs a tag?

manics: Looks like it was already tagged, updated.

PR #60, opened 2017-10-19
Pin version of flannel playbook in roles/kubernetes (IDR-0.4.4)

manics:
An upstream change broke something. The longer term plan is to move to Kubespray for deploying ansible instead of maintaining our own role.


joshmoore:
:+1:

PR #61, opened 2017-10-19
Remove nfs group creation (IDR-0.4.4)

manics:
Minor cleanup- this is already handled elsewhere


joshmoore:
:100: for less redundancy.

PR #62, opened 2017-10-19
Bump requirements (IDR-0.4.4)

manics:
Includes breaking changes to `openmicroscopy.postgresql`

PR #63, opened 2017-10-19
Bump OMERO, update PG config (IDR-0.4.4)

manics:
Additional cleanups include moving omero-web into a separate playbook

~~--depends-on https://github.com/IDR/deployment/pull/62~~ Commits from https://github.com/IDR/deployment/pull/62 are now included here

### DIFF
--- a/ansible/group_vars/database-hosts.yml
+++ b/ansible/group_vars/database-hosts.yml
@@ -29,3 +29,18 @@ idr_omero_readonly_database:
   user: omeroreadonly
   password: "{{ idr_secret_postgresql_password_ro | default('omero') }}"
   statement_timeout: 60000
+
+postgresql_install_extensions: True
+
+# TODO: add a variable to easily enable/disable
+# TESTING ONLY!
+#postgresql_server_conf:
+#pg_stat_staments?
+#  shared_preload_libraries: "'pg_stat_statements'"
+#  pg_stat_statements.max: 20000
+#Database tuning?
+#  default_statistics_target = 1000
+#  maintenance_work_mem = 256MB
+#  checkpoint_completion_target = 0.9
+#  effective_cache_size = "{{ (ansible_memtotal_mb / 2) | int }}MB"
+#  work_mem = 64MB
--- a/ansible/group_vars/database-hosts.yml
+++ b/ansible/group_vars/database-hosts.yml
@@ -10,7 +10,11 @@ postgresql_server_auth:
   user: "{{ idr_omero_readonly_database.user }}"
   address: 0.0.0.0/0
 
-postgresql_users_databases:
+postgresql_databases:
+- name: idr
+  owner: omero
+  restrict: True
+postgresql_users:
 - user: omero
   password: "{{ idr_secret_postgresql_password | default('omero') }}"
   databases: [idr]
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -6,7 +6,7 @@
 ######################################################################
 # Shared variables
 
-idr_omero_release: "0.4.2-2"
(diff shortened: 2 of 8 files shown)


ansible/group_vars/database-hosts.yml
```diff
@@ -29,3 +33,18 @@ idr_omero_readonly_database:
   user: omeroreadonly
   password: "{{ idr_secret_postgresql_password_ro | default('omero') }}"
   statement_timeout: 60000
+
+postgresql_install_extensions: True
+
+# TODO: add a variable to easily enable/disable
```
joshmoore: Is this worth capturing outside of idr/deployment?

ansible/requirements.yml
```diff
@@ -176,3 +176,7 @@
 - name: openmicroscopy.prometheus-node
   src: https://github.com/openmicroscopy/ansible-role-prometheus-node/archive/0.1.1.tar.gz
   version: 0.1.1
+
+- name: openmicroscopy.prometheus-postgres
+  src: https://github.com/openmicroscopy/ansible-role-prometheus-postgres/archive/0.1.0.tar.gz
```
joshmoore: Move to galaxy?

ansible/requirements.yml
```diff
@@ -100,7 +100,7 @@
   version: 1.1.0
 
 - src: openmicroscopy.postgresql
-  version: 2.0.0
+  version: 3.0.0-m2
```
joshmoore: Since this is the first production IDR version with these, are we getting close to dropping `m` then?

ansible/group_vars/omero-hosts.yml
```diff
@@ -83,6 +83,20 @@ omero_user_system: omero-server
 ######################################################################
 # openmicroscopy.omero-web role vars
 
+idr_omero_web_public_url_filters_webclient_exclude:
```
joshmoore: Is this something for the role itself?

joshmoore:
Few questions, but otherwise, this looks good to go.

ansible/group_vars/database-hosts.yml
manics: I see this as still in the experimental phase (which is why it's commented out), but if it turns out to be useful we might just enable it unconditionally.

ansible/requirements.yml
manics: It depends how long we need to run this in IDR production before we decide it's good enough for general use, but eventually yes.

ansible/requirements.yml
manics: 👍  AFAIK no-ones complained about https://github.com/openmicroscopy/ansible-role-postgresql/tree/3.0.0-m2

ansible/group_vars/omero-hosts.yml
manics: I don't know. Do we need a filter with `https://docs.openmicroscopy.org/omero/5.4.0/sysadmins/config.html#std:property-omero.web.public.get_only`?

ansible/requirements.yml
sbesson: Now we've deployed non milestone versions, should we move this to `3.0.0` and omero-server to `2.0.1`?

manics: Not now, it's already deployed. Next release?

sbesson: Given the effective difference between `3.0.0-m2` and `3.0.1`, agreed I would not block IDR 0.4.4 for this.

Updating our requirements to consume the latest non-milestone role tags before we spin up a new idr-testing server would certainly be valuable.

sbesson:
Questions and todos captured in https://trello.com/c/acEYekZX/80-idr-044-deployment-todos. Merging.

PR #64, opened 2017-10-19
Ensure journald logs are maintained after a reboot (IDR-0.4.4)

manics:
The CentOS 7 cloud image default to not keeping logs over a reboot. See https://unix.stackexchange.com/a/159390


joshmoore:
Makes sense. Cheers.

PR #65, opened 2017-10-19
Prometheus updates (IDR-0.4.4)

manics:
Adds monitoring of Postgres, and some clean-ups.


sbesson:
All makes sense. Merging.

PR #66, opened 2017-10-19
Update about page, link to grafana images (IDR-0.4.4)

manics:
This adds optional vars for showing the version and other metadata on the about page, and adds grafana dashboard images (since these are not interactive it should be secure enough, see https://github.com/IDR/deployment/pull/58).


joshmoore:
Should http://idr/grafana currently load?

manics:
No, only the full URLs will work.

joshmoore:
Understood.

PR #67, opened 2017-10-19
Update scripts for latest deployment structure (IDR-0.4.4)

manics:
Everything is now in a single network. This also adds a warning to the delete script when deleting non-test deployments.


joshmoore:
:+1:

PR #72, opened 2017-12-13
Increase application_server.max_requests from 300 to 30000 (IDR-0.4.5)

manics:
Our docs currently mention `omero.web.application_server.max_requests 500` without any explanation of why, or how to decide: https://docs.openmicroscopy.org/omero/5.4.0/sysadmins/unix/install-web/web-deployment.html#gunicorn-configuration

A slack discussion suggested it's not needed https://openmicroscopy.slack.com/archives/C0K5EED3R/p1507894075000132

Cards on our sysadmin board suggest web needs regular restarts: https://trello.com/c/DC5CB4Ty/34-bug-omeroweb-processes-threads

This increases the limit from 300 to 30000 in the hope that it's not really needed so having a large value won't matter, but that it's low enough so that if there are problems web workers will be restarted before they occur.

Related card:
- https://trello.com/c/n6JhjbPu/17-poor-performance-with-django-prometheus-after-several-weeks this is a disadvantage of the django-prometheus collector, but it did highlight the low value of `omero.web.application_server.max_requests`

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -106,7 +106,7 @@ omero_web_omego_additional_args: "{{ idr_omero_omego_additional_args }}"
 omero_web_config_set:
   # web
   omero.web.application_server: wsgi-tcp
-  omero.web.application_server.max_requests: 300
+  omero.web.application_server.max_requests: 30000
   omero.web.wsgi_workers: 25
   #omero.web.wsgi_timeout 30
   #omero.web.wsgi_args -- '--forwarded-allow-ips=YOUR_IP'


jburel:
I assume we will have to keep monitoring to see if it is an acceptable value
Will you open a doc PR to remove the configuration?

manics:
>  Will you open a doc PR to remove the configuration?

No, I don't understand it and there seem to be conflicting views. The doc needs to be reviewed by someone who understands what's going on, and if necessary it could be removed, extended with more explanation, or moved to the production/hardening section of the OMERO sysadmin docs.

will-moore:
Was asked to look at this PR, but I'm afraid I have no experience of this config.

joshmoore:
Think for all of us, this implies some research.

manics:
--exclude

sbesson:
Now deployed together a build included with https://github.com/openmicroscopy/openmicroscopy/pull/5638 on idr testing server.

PR #82, opened 2018-01-29
Increase number of omeroreadonly from 2 to 4 (IDR-0.4.6)

manics:
Self explanatory.


sbesson:
While testing the redeployment of `test45`, realized that `scripts/os-idr-delete.sh` should be updated accordingly.

manics:
Updated

manics:
I guess this should be pushed to 0.4.6? Then we can also test reducing the resource allocations to the read-only instances.

manics:
Rebased, and cut down the size of the omero-readonly servers to 32GB.

PR #85, opened 2018-02-08
Exclude nginx "a client request body is buffered ... file" alerts (IDR-0.4.5)

manics:
To be tested on idr-testing


manics:
Now ignoring `buffered to a temporary file`

PR #87, opened 2018-02-23
Change min-width of idr-sessions grafana panels (IDR-0.4.6)

manics:
Needed to fit in more omeroreadonly servers.

Also change initial display to 1h.


sbesson:
Renders as

screen shot 2018-02-28 at 08 48 56

vs currently

screen shot 2018-02-28 at 08 51 27

LGTM

PR #88, opened 2018-02-27
Use centos username rather than id

sbesson:
With Ansible 2.4.3.0, this change was necessary to run the playbook. Without
it, the download task fails with an error of type
`sudo: unknown user: 1000\r\nsudo: unable to initialize policy plugin\r\n`


manics:
Can you add a comment that the UID must match the user in https://hub.docker.com/r/imagedata/mineotaur/ since the docker container requires write access

sbesson:
@manics pushed comment and restarted Travis

PR #73, opened 2017-12-13
Load balance OMERO (part two) (IDR-0.4.5)

manics:
This is the other main part of the proposal merged in https://github.com/bramalingam/idr-py/pull/1

See the comment at the opt of https://github.com/IDR/deployment/compare/master...manics:idr045-haproxyomero?expand=1#diff-a9e374e386631a6af421a846f80737fb for an explanation

This is not usable from outside yet, as it requires addition ports to be opened in the security group.

PR #89, opened 2018-02-27
K8s idr-omero-public is now a proxy config server (IDR-0.4.6)

manics:
This changes the internal kubernetes alias for the omero-readonly servers to point to the internal IP of the proxy instead, from where the  client config file can be downloaded.


manics:
I tried to separate out this change, but it's interlinked with https://github.com/IDR/deployment/pull/73

PR #90, opened 2018-02-27
Open additional external ports for OMERO (IDR-0.4.6)

manics:
Uses dev version of openstack-idr-security-groups

- [x] --depends-on https://github.com/IDR/ansible-role-openstack-idr-security-groups/pull/3

Security groups are shared so this will open up ports for all deployments


sbesson:
Upstream role has been release as `1.2.0` and deployed to Galaxy.

manics:
Updated

PR #95, opened 2018-03-12
Add provisioning and deployment playbooks for idr-ftp

manics:
Proposed new idr-ftp deployment.

Changes: https://github.com/IDR/deployment/pull/95/commits/c37e1b7e287e375ad740bce16b5faf1567e83d94 makes it easier to use flags to provision a subset of the IDR VMs.
- renames openstack-create-infrastructure.yml to openstack-create-publicidr.yml (alternative name suggestions welcome). Git tracking fails to track this as a rename with minor changes so it looks like a big change.
- moves the default provisioning vars at the top of openstack-create-infrastructure.yml to vars/openstack-vars.yml
- openstack-create-infrastructure.yml becomes a parent playbook with flags to disable parts of the whole IDR (default: enable everything)

Additions:
- FTP provisioning playbook openstack-create-ftp.yml
- FTP deployment playbook idr-ftp.yml

This should be deployed in the same way as the rest of the IDR.

- [x] --depends-on https://github.com/manics/ansible-role-anonymous-ftp


ansible-ftp/idr-ftp.yml
```diff
@@ -0,0 +1,93 @@
...
+      name: docker-python
+      state: present
+
+  - name: Run docker vsftpd
+    become: yes
+    docker_container:
+      image: openmicroscopy/vsftpd-anonymous-upload:0.1.0
+      name: vsftpd
+      published_ports:
+      - "32021:21"
```
manics: This matches the current deployment, but we could change it to `21:21` so that you don't need to specify the port when connecting

ansible-ftp/idr-ftp.yml
```diff
@@ -0,0 +1,93 @@
...
+      - "32031:32031"
+      state: started
+      restart_policy: always
+      volumes:
+      - /data/idrftp/incoming:/var/lib/ftp/incoming
+      - /etc/vsftpd:/etc/vsftpd:ro
+
+  vars:
+    #idrftp_public_address: 193.62.55.107
+    idrftp_public_address: 10.0.51.177
```
manics: This is an internal test address.

manics:
If someone comes up with a better name I'll change it 😀

sbesson:
Briefly reviewed with @manics. Couple of immediate thoughts:

- although combining `publicidr` + `ftp` (+ `vae`) into a single deployment is a possibility, we might not have everything enabled by default (i.e. matching how we would be deploying it)
- also I'd like to have a better name than `public_idr`. Immediately, I propose to get this in and adjust the private variables wherever necessary.

sbesson:
Merging as the current state of the deployment. Discussed today, an alternative to `public_idr` might be `core`.

PR #99, opened 2018-03-19
Haproxy separate TCP logs (IDR-0.4.7)

manics:
This directs the incoming TCP connection logs to a separate file `/var/log/haproxy/haproxy-tcp.log`. A few non-connection logs will still creep in but these can be ignored.


joshmoore:
Changes looks as expected. :+1: Once green, I'll merge for inclusion in prod47, and if there are any issues, we'll have to open a follow-up PR.

PR #101, opened 2018-03-20
Force hostname in kubeadm init/join

manics:
@joshmoore ran into a problem where a node couldn't join itself due to a mismatch between hostnames, despite this having worked fine for @sbesson during the last deployment.


joshmoore:
I'm unsure why travis is unhappy here, but running with the change was certainly more successful than without. Merging and we'll see what the mainline travis looks like.

manics:
Travis seems happy with the merge https://travis-ci.org/IDR/deployment/builds/356060585 ¯\_(ツ)_/¯

joshmoore:
:+1:

PR #103, opened 2018-03-26
idr050: modifications for OMERO 5.4

joshmoore:
see: https://github.com/openmicroscopy/openmicroscopy/pull/5684

### DIFF
--- a/ansible/idr-omero-readonly.yml
+++ b/ansible/idr-omero-readonly.yml
@@ -79,16 +79,13 @@
       recurse: yes
       state: directory
 
-  # TODO: Rendering in web is current broken with omero.cluster.read_only true
   - name: OMERO.server read-only configuration
     become: yes
     copy:
       content: |
         config set -- omero.data.dir /data/OMERO
-        #config set -- omero.cluster.read_only true
-        config set -- omero.cluster.node_provider ome.security.basic.InMemoryNodeProvider
-        config set -- omero.sessions.session_manager ome.services.sessions.InMemorySessionManagerImpl
-        config set -- omero.security.event_provider ome.security.basic.InMemoryEventProvider
+        config set -- omero.cluster.read_only true
+        config set -- omero.pixeldata.memoizer.dir.local /tmp/BioFormatsCache
       dest: /opt/omero/server/config/omero-readonly-config.omero
       force: yes
     notify:
@@ -105,7 +102,6 @@
       state: directory
     with_items:
     - /data/OMERO
-    - /data/OMERO/ManagedRepository
 
   # Note BioFormatsCache is a separate read-write shared cache
   - name: OMERO.server read-only data dir symlinks
@@ -118,11 +114,10 @@
     with_items:
     - Files
     - FullText
-# TODO: OMERO read-only writes to .omero and ManagedRepository/.omero
-#    - ManagedRepository
+    - ManagedRepository
     - Pixels
     - Thumbnails
-#    - .omero
+    - .omero
 
   - name: OMERO.server BioFormatsCache symlinks
     become: yes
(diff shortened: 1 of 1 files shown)


ansible/idr-centos-init.yml
```diff
@@ -4,7 +4,16 @@
 - hosts: "{{ idr_environment | default('idr') }}-hosts"
 
   tasks:
-  - name: Create persistent journald directory
+  - name: centos-init | Install locate
+    become: yes
+    yum:
+      name: mlocate
```
manics: What's the background for this? The PR openmicroscopy/openmicroscopy#5684 in the description doesn't mention it.

ansible/idr-omero-readonly.yml
```diff
@@ -79,16 +79,13 @@
...
     become: yes
     copy:
       content: |
         config set -- omero.data.dir /data/OMERO
-        #config set -- omero.cluster.read_only true
-        config set -- omero.cluster.node_provider ome.security.basic.InMemoryNodeProvider
-        config set -- omero.sessions.session_manager ome.services.sessions.InMemorySessionManagerImpl
-        config set -- omero.security.event_provider ome.security.basic.InMemoryEventProvider
+        config set -- omero.cluster.read_only true
+        config set -- omero.pixeldata.memoizer.dir.local /tmp/BioFormatsCache
```
manics: Does this mean the NFS share https://github.com/IDR/deployment/blob/IDR-0.4.7-1/ansible/idr-omero-readonly.yml#L24 can either be removed or made read-only?

ansible/idr-centos-init.yml
```diff
@@ -4,7 +4,16 @@
...
+      state: present
+
+  - name: centos-init | Update locate db
+    command: updatedb
```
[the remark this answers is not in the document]
joshmoore: I hadn't added previously. @mtbc : was the behavior on idr-experimental ok for you? If so, I could drop. Otherwise, the only thing that occurs to me is to setup in cron.

ansible/idr-centos-init.yml
joshmoore: It's not an OMERO-specific request but an "IDR devops" request. i.e. for the likes of Mark to better find things on these servers.

ansible/idr-omero-readonly.yml
joshmoore: It _could_ be made read-only, but this is actually one of the places where I think we've gone too far. Specifically, I *want* to be able to write to the global BioFormatsCache from the read-only servers. I don't need to _yet_ since we're not bumping our Bio-Formats version, but we will need to _soon_.

ansible/idr-centos-init.yml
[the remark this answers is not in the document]
mtbc: I did an updatedb manually there. Waiting for `crond` to run `/etc/cron.daily/mlocate` would be fine.

ansible/idr-centos-init.yml
manics: From a "devops" perspective we shouldn't really need this since it means we haven't defined/documented where everything is. Does it need to be on every server, or just the OMERO ones?

joshmoore: OMERO and Database was the initial requirement.

manics: Can you make it specific to those then? `idr-00-preinstall.yml` is run by everything- publicidr, ftp, and the k8s vae.

joshmoore:
Relaunched travis with gh-109

ansible/idr-centos-init.yml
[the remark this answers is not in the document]
mtbc: mlocate's significantly better than predecessors in my limited experience.

ansible/idr-centos-init.yml
joshmoore: Added `idr_debug` after discussion in `#idr`.

Note: also dropped the cron section since @mtbc pointed out that a `daily` job is installed by default.

ansible/idr-centos-init.yml
```diff
@@ -4,7 +4,14 @@
 - hosts: "{{ idr_environment | default('idr') }}-hosts"
 
   tasks:
-  - name: Create persistent journald directory
+  - name: centos-init | Install locate
+    become: yes
+    yum:
+      name: mlocate
+      state: present
+    when: idr_debug | default(false) | bool
```
manics: Optional: If it's going to make things easier in future set a default in https://github.com/IDR/deployment/blob/IDR-0.4.7-1/ansible/group_vars/all.yml

ansible/openstack-create-publicidr.yml
```diff
@@ -120,42 +120,6 @@
...
-    # Analysis Volumes
-
-    - role: openmicroscopy.openstack-volume-storage
-      openstack_volume_size: 100
-      openstack_volume_vmname: "{{ idr_environment_idr }}-dockermanager"
-      openstack_volume_name: data
-      openstack_volume_device: /dev/vdb
-      openstack_volume_source: "{{ idr_volume_dockermanager_data_src }}"
-
-
```
manics: This is fine for now, but we'll need to clean up all references to these servers in idr-proxy at some point.

ansible/idr-centos-init.yml
joshmoore: I thought about that, but since I haven't rolled it out elsewhere, I think I'll leave this as "internal API" for the moment. But agreed, having a section in `all.yml` with "expert" variables will be useful.

ansible/openstack-create-publicidr.yml
joshmoore: Agreed. I couldn't tell where I'd conflict with your work, so tried to go for the simplest thing possible.

joshmoore:
From my point-of-view as soon as all 0.4.8 is merged, this can be merged.

ansible/idr-omero-readonly.yml
joshmoore: @mtbc suggests in https://trello.com/c/q5L39XPZ/5-bioformatscache-is-expecting-read-write that if I omit this line, then the old behavior will be used. Likely worth testing.

ansible/idr-centos-init.yml
manics: When we move to Ansible 2.5 we can use the new `never` tag instead http://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html

ansible/idr-omero-readonly.yml
joshmoore: Pushed a removal of `omero.pixeldata.memoizer.dir.local`

sbesson:
:+1: What are the remaining steps to getting this merged and deployed on `prod50`?

joshmoore:
Happy to run on prod50 today from a merged stated if there are no objections.

PR #106, opened 2018-04-11
Workaround for incompatibilty between Docker 1.13.1 and Kubernetes 1.7.4

manics:
Since these servers will go away in the next release this is a quick fix.


manics:
I think the travis failure is due to https://github.com/openmicroscopy/openmicroscopy/pull/5674 not being on metadata53

manics:
That PR isn't the problem. The issue is `omero-marshal==0.5.1` fails in a clean virtualenv on CentOS 7:
```
virtualenv m53-test
m53-test/bin/pip install omero-marshal==0.5.1
```

joshmoore:
Thanks, @manics. Merging and tagging this as `0.4.7-1`. For the failing pip issue, we may then need a `0.4.7-2` if we want this repo to be deployable.

PR #107, opened 2018-04-16
Minor idempotence correction for journald dir

manics:
When journald starts it resets the permissions on `/var/log/journal`. This PR ensures the perms set by ansible match those set by journald to avoid idempotence warnings when rerunning after journald has been restarted for other reasons.


joshmoore:
Relaunched travis with gh-109

joshmoore:
Now green on travis with gh-109 and also successfully run locally as prod50 with a merged build. :+1:

PR #109, opened 2018-04-17
Bump to OMERO 0.4.7

joshmoore:
This is an emergency release of OMERO to fix
failing pip installations of omero-marshal.

see: https://github.com/openmicroscopy/omero-marshal/pull/50

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -6,7 +6,7 @@
 ######################################################################
 # Shared variables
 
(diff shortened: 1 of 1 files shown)


joshmoore:
Merging so we can get travis green. I can see tagging this as `IDR-0.4.7-2` or even having a more formal `IDR-0.4.8` to match the state of all @manics PRs.

sbesson:
Given the amount of work and the fact many of these are not on `prod47`, I would personally vote for calling it `IDR-0.4.8` even if we do not plan to deploy it.

PR #110, opened 2018-04-19
Update docs

manics:
Update provisioning and deployment docs. Also add note on viewing Grafana.

Closes https://github.com/IDR/deployment/issues/105

PR #112, opened 2018-04-20
Block proxy /webclient/login

manics:
This blocks the OMERO.web login page on the front-end proxy to prevent private state from being cached. See https://github.com/IDR/deployment/pull/110#discussion_r183026496

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -224,6 +224,9 @@ nginx_proxy_direct_locations:
   redirect302: /webclient/?experimenter=-1
 - location: "= /"
   redirect302: /about
+  # Block webclient login
+- location: "^~ /webclient/login"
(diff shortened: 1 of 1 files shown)


joshmoore:
Applied to omeroreadonly on experimental. The behavior of nginx (with caching) is:

 * manually go to `/login` redirect to data as public user
 * manually go to `/logout`, confirm, redirect to data as the public user
 * click on "logout", the same.

I haven't been able to find a sequence which doesn't end up showing me anything but data as the default public user setup. The only issue I could think of is if we _want_ to have login, but I don't think we do. Certainly not in production.

If there are no objections tomorrow, I'll apply to prod50.

joshmoore:
General sign-off during the SA meeting. Anyone needing login will need to port-forward to the backend server. I'll deploy this to prod50 now independent of a tag. Merging.

PR #114, opened 2018-05-02
WIP: Test omeroreadonly in molecule

manics:
This changes the NFS exports on omeroreadwrite to be relative to `/data` instead of `/` by setting `fsid=0`. The primary benefit of this is it means omero readonly should be testable with molecule docker. Previously it was disabled as the OMERO data dir couldn't be mounted on the read-only servers in docker due to the docker filesystem type. Making `omeroreadwrite:/data` an external docker volume in molecule gets around this problem.

The server side data volumes are unchanged, this only affects the client-side mounts on the read-only servers. Since prod50 is already deployed we should save this for test51/prod51. However, it should be possible to test this locally alongside https://github.com/IDR/deployment/pull/103

### DIFF
--- a/ansible/idr-01-install-idr.yml
+++ b/ansible/idr-01-install-idr.yml
@@ -9,9 +9,6 @@
 - include: idr-omero-web.yml
 
 - include: idr-omero-readonly.yml
-  tags:
-  # Requires NFS shares which can't be configured in Docker
-  - skip_if_molecule_docker
 
 - include: idr-docker.yml
 
--- a/ansible/idr-omero-readonly.yml
+++ b/ansible/idr-omero-readonly.yml
@@ -188,5 +188,7 @@
 
   roles:
   - role: openmicroscopy.omero-server
+    when: not (TRAVIS | default(False))
+
 
   environment: "{{ idr_ANSIBLE_ENVIRONMENT_VARIABLES | default({}) }}"
--- a/ansible/idr-omero-readonly.yml
+++ b/ansible/idr-omero-readonly.yml
@@ -18,6 +18,10 @@
 
   - role: openmicroscopy.nfs-share
     nfs_shares:
+      # WARNING: This makes /data the root for all exports from this server
+      /data:
+      - host: "*"
+        options: 'ro,fsid=0'
       /data/OMERO:
       - host: "*"
         options: 'ro'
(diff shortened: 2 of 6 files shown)


manics:
```
ERROR: Idempotence test failed because of the following tasks:
ERROR: * [idr-omeroreadonly-docker] => OMERO.server read-only copy read-write repo uuid
```
Presumably this should pass with #103 since it is no longer necessary to copy the repo uuid: https://github.com/IDR/deployment/pull/103/files#diff-b5266c3d9292cae6c983f2875a8a8e0cL152

joshmoore:
@manics : do you want me to cherry-pick anything?

manics:
No, the only way to get this tested with docker molecule is to change the client NFS mount config. This shouldn't conflict with any of your changes since you're not modifying the NFS mounts.

manics:
I've rebased. I suspect travis will still fail due to resource issues (running two OMERO servers requires a lot of RAM, Travis has 7.5GB available https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System).

If this works locally then we can decide what to do.

manics:
I've got this passing on Docker locally after increasing the Docker on Mac Memory. I set it to 12 GB, `docker stats` suggests it needs around 9.5GB to run the full test:
```
CONTAINER ID        NAME                        CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
28cc2b88a9b8        idr-proxy-docker            0.01%               77.98MiB / 11.71GiB   0.65%               104MB / 16.4MB      158MB / 519MB       15
31a6f3390c6e        idr-omeroreadonly-docker    2.97%               4.554GiB / 11.71GiB   38.88%              512MB / 59.5MB      19MB / 568MB        235
a4414f9a5406        idr-omeroreadwrite-docker   2.98%               4.701GiB / 11.71GiB   40.14%              555MB / 117MB       14.2MB / 675MB      245
bc7b3d7583d7        idr-database-docker         0.46%               65.17MiB / 11.71GiB   0.54%               132MB / 97.4MB      22MB / 883MB        17
```

Since this exceeds the resources provided by travis one option is to make the readonly playbook conditional on being run outside travis, either locally in development or on a (future) CI system. Thoughts?

sbesson:
As this was still included in the daily integration branch, this PR was deployed on `test51`. As noticed this morning, this had some severe performance impact when accessing the image data as the `/data/OMERO/BioFormatsCache` directory was effectively mounted read-only which prevents Bio-Formats cache files from being loaded with the current software.

Redeploying without this PR sufficed to restore the previous behavior (after fixing manually a couple of `/data/` subfolders). Excluded the PR for now.

sbesson:
Closing as per the conflicts

PR #115, opened 2018-05-07
Add .html suffix to mineotaur redirect

joshmoore:
see: https://github.com/IDR/deployment/pull/113

(Manually applied to prod50)


joshmoore:
Would suggest getting this into the 0.5.0 tag.

joshmoore:
A general dropping of `.html` across the board would be my long-term preference, but for the moment, we're `.html` everywhere jekyll related.

PR #118, opened 2018-06-13
Update default FTP size and allow it to be configurable

sbesson:
Follow-up of #117, this addition was used when recreating the IDR FTP service to match the recent extension of the volume.

The default is bumped to 15TB and a new `idr_environment_ftp_size` variable is added to override it via Ansible variables. Alternatively, we could set a much smaller default in the public playbook e.g. 100GB and move the production idr-ftp value to the host variable.

### DIFF
--- a/ansible/openstack-create-ftp.yml
+++ b/ansible/openstack-create-ftp.yml
@@ -76,7 +76,7 @@
   # Volumes
 
   - role: openmicroscopy.openstack-volume-storage
-    openstack_volume_size: 5000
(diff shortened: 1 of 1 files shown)


sbesson:
The variable is currently unused but if it is useful, I could certainly add it to the current production FTP private variables i.e. 

```
idr_environment_ftp_size: 15000
```
If we want to override, I would also vote for lowering the default value to 10GB.

manics:
Let's just leave it for now then.

joshmoore:
> If we want to override, I would also vote for lowering the default value to 10GB.

:+1:

PR #119, opened 2018-06-15
Make idr-management:/data into a volume (IDR-0.5.1)

manics:
We're generating a lot of logs and metrics (or there's a misconfiguration which means the  elasticsearch logs aren't being deleted after 14 days). This should ensure they're stored on a 100GB volume instead of in the root partition.

https://trello.com/c/PmYJjRO1/45-make-idr-management-data-into-a-volume

Untested.


manics:
Deployed on test51

sbesson:
Thanks. Merging and we can check on `prod51` that everything works as expected.

PR #120, opened 2018-06-15
Prometheus monitoring bump (IDR-0.5.1)

manics:
Bump all prometheus exporters, prometheus server, and grafana.

### DIFF
--- a/ansible/management-grafana.yml
+++ b/ansible/management-grafana.yml
@@ -9,19 +9,21 @@
   - name: Run docker grafana
     become: yes
     docker_container:
-      image: grafana/grafana
+      image: grafana/grafana:5.1.3
       env:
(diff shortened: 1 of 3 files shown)


manics:
Deployed on test51

manics:
> if we will eventually need a high-level idr_monitoring_enabled flag

What are you thinking? I can't see any disadvantages to enabling monitoring since it's all self-contained within the IDR deployment. For our central (OME) monitoring server we're free to include or not include the metrics for individual IDR deployments based on the external proxy name (idr, idr-testing, idr-next, etc).

joshmoore:
Probably a question of how large the monitoring overhead becomes. i.e. as it starts to become a proper cluster we may need to re-evaluate.

PR #121, opened 2018-06-15
Allow access to prometheus federation endpoint (currently no auth) (IDR-0.5.1)

manics:
This allows access to the prometheus federation endpoint.
To be decided:
- [x] Authenticate or leave open? [Decision at IDR meeting 2018-06-18: no auth]


manics:
Deployed on test50 and test51

joshmoore:
> [Decision at IDR meeting 2018-06-18: no auth]

Expanding on that, to be reconsidered when/if we begin allowing individual submitters to login since their login information would become public.

PR #123, opened 2018-06-21
rsync download is on omeroreadwrite host

manics:
The rsync data source is only on the read-write omero server.


joshmoore:
Makes sense to me. (I don't even know if HA rsync is a thing).

PR #127, opened 2018-07-09
Add configurable NFS mounts for internal read-only access to raw data

manics:
See:
- https://trello.com/c/GKEgLjPb/51-request-nfs-biostudies-and-mount

### DIFF
--- a/ansible/idr-ebi-nfs.yml
+++ b/ansible/idr-ebi-nfs.yml
@@ -4,15 +4,16 @@
 
   tasks:
 
-  - name: Mount NFS share
+  - name: Mount NFS shares
     become: yes
     mount:
       fstype: nfs
-      name: "{{ nfs_mountpoint }}"
+      name: "{{ item.dst }}"
       # Removed rsize=8192,wsize=8192, this should be auto-negotiated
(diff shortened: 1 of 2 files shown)


ansible/idr-ebi-nfs.yml
```diff
@@ -36,6 +37,5 @@
 
 
   vars:
-    nfs_mountpoint: /nfs/bioimage
     nfs_filesets_dir: /nfs/bioimage/drop/complete
```
sbesson: What would you do about this variable? In the absence of `nfs_mountpoint` it still feels pretty specific to the current IDR deployment.

manics: It is very specific, but I couldn't immediately see how it fits together with https://github.com/openmicroscopy/management_tools/blob/master/idr/playbooks/idr-links.yml so left it as it is for now

manics: For reasons lost to the distant past these seem to be redundant. I've removed them.

sbesson:
Thanks @manics. Deployed on `test51` and the expected shares are now correctly mounted. I can confirm 41c587d has no impact on the deployment either.

PR #129, opened 2018-08-08
Use new deploy_archive role for deploying idr website

manics:
Replace jekyll build with a pre-built artefact containing the IDR `/about` pages.

Due to the changes required in https://github.com/IDR/idr.openmicroscopy.org this is currently using a temporary build http://users.openmicroscopy.org.uk/~spli/idr/idr-www-test/idr.openmicroscopy.org.tar.gz

Before deploying this manually run `cd /srv/www; sudo mv html html.old && sudo ln -s html.old html`

- [x] --depends-on https://github.com/openmicroscopy/ansible-role-deploy-archive/pull/2 for a selinux fix (I haven't bumped requirements yet because this doesn't affect travis/docker tests).

### DIFF
--- a/ansible/idr-proxy-about.yml
+++ b/ansible/idr-proxy-about.yml
@@ -2,36 +2,14 @@
 
 - hosts: "{{ idr_environment | default('idr') }}-proxy-hosts"
 
-  pre_tasks:
-
-  # This seems complicated because the config file needs to exist before
-  # jekyll-build, but we also want to notify jekyll-build of changes which
-  # can't be done until after the jekyll-build role
-
-  - name: create jekyll config directory
-    become: yes
-    file:
-      path: "{{ jekyll_build_config.0 | dirname }}"
-      recurse: yes
-      state: directory
-
-  - name: create dummy jekyll config
-    become: yes
-    copy:
-      content: ''
-      dest: "{{ jekyll_build_config.0 }}"
-      # Only create if it doesn't exist
-      force: no
-
   roles:
-  - role: openmicroscopy.jekyll-build
+  - role: openmicroscopy.deploy_archive
+    become: yes
 
   tasks:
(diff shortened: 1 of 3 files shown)


ansible/idr-proxy-about.yml
```diff
@@ -2,36 +2,14 @@
...
+    become: yes
 
   tasks:
-  - name: create jekyll config
+  - name: Set website displayed version
     become: yes
     copy:
-      content: "{{ idr_openmicroscopy_org_config | to_nice_yaml }}"
-      dest: "{{ jekyll_build_config.0 }}"
+      content: "{{ idr_openmicroscopy_org_config.idr_deployment_version }}"
```
sbesson: Do we still need a dictionary structure with one item (esp. now the deployment date is replaced by the build date and moved upstream) or should this be reduced to a single `idr_deployment_version` variable

manics: No reason other than to minimise changes. Potentially we could change `VERSION` into `version.json` and add a lot more metadata about the IDR (probably heading into bioschemas territory 😊), but I'm happy to flatten it to a single var.

manics:
Bumped the role and renamed the version variables to hopefully disambiguate between the website version and the IDR version.

manics:
Rested on idr-testing by deleting `html`, deploying the website with the old jekyll role, then running
```
cd /srv/www; sudo mv html html.old && sudo ln -s html.old html
```
and redeploying using this updated PR and my own merge build of idr.openmicroscopy.org .

manics:
Updated to use https://github.com/IDR/idr.openmicroscopy.org/releases/tag/2018.08.09, deployed to test51.

jburel:
Strategy used to update idr.o.o today as part of the vae release.

PR #130, opened 2018-08-09
Shared buffers bump

sbesson:
See https://trello.com/b/AM7jhXRQ/idr-051

As tested during the IDR 0.5.0 by @mtbc, this PR bumps the value of `shared_buffers` to 1GB using the `postgresql_server_conf` dictionary. Additionally it bumps `openmicroscopy.postgresql` to the latest version of the role. This should be harmless as there has been no change in the logic in the role, only documentation - see https://github.com/openmicroscopy/ansible-role-postgresql/compare/3.0.0-m2...3.0.1.

### DIFF
--- a/ansible/group_vars/database-hosts.yml
+++ b/ansible/group_vars/database-hosts.yml
@@ -36,15 +36,5 @@ idr_omero_readonly_database:
 
 postgresql_install_extensions: True
 
-# TODO: add a variable to easily enable/disable
-# TESTING ONLY!
-#postgresql_server_conf:
-#pg_stat_staments?
-#  shared_preload_libraries: "'pg_stat_statements'"
-#  pg_stat_statements.max: 20000
-#Database tuning?
-#  default_statistics_target = 1000
-#  maintenance_work_mem = 256MB
-#  checkpoint_completion_target = 0.9
-#  effective_cache_size = "{{ (ansible_memtotal_mb / 2) | int }}MB"
-#  work_mem = 64MB
+postgresql_server_conf:
+  shared_buffers: 1GB
--- a/ansible/group_vars/database-hosts.yml
+++ b/ansible/group_vars/database-hosts.yml
@@ -37,4 +37,4 @@ idr_omero_readonly_database:
 postgresql_install_extensions: True
 
 postgresql_server_conf:
-  shared_buffers: 1GB
+  shared_buffers: "{{ (ansible_memtotal_mb / 4) | int }}MB"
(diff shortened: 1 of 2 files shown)


ansible/group_vars/database-hosts.yml
```diff
@@ -36,15 +36,5 @@ idr_omero_readonly_database:
...
-#  shared_preload_libraries: "'pg_stat_statements'"
-#  pg_stat_statements.max: 20000
-#Database tuning?
-#  default_statistics_target = 1000
-#  maintenance_work_mem = 256MB
-#  checkpoint_completion_target = 0.9
-#  effective_cache_size = "{{ (ansible_memtotal_mb / 2) | int }}MB"
-#  work_mem = 64MB
+postgresql_server_conf:
+  shared_buffers: 1GB
```
manics: Any chance of making this a percentage of total mem instead to make it easier to test on smaller systems?

joshmoore: e.g. https://github.com/ANXS/postgresql/pull/235/files#diff-0ab2a3945984f21ca36ce37d625d716cR104

sbesson: You mean using `ansible_memtotal_mb` ?

mtbc: I wonder if there's rather more risk of allocating uselessly much? Are such tiny systems a concern?

joshmoore: I would think with a min & max clause.

sbesson: Fine with me. I followed @mtbc's initial investigation suggesting 1GB is sufficient for our needs (https://trello.com/c/4V4lqQre/493-mapr-is-slow). https://www.postgresql.org/docs/9.6/static/runtime-config-resource.html seems to point towards 25% as a reasonable default. I can switch to that unless someone proposes an another percentage

sbesson:
Deployed on `test51`

```diff
TASK [openmicroscopy.postgresql : postgres | postgresql config file] ***************************************************************************************************************************************************
--- before: /var/lib/pgsql/9.6/data/postgresql.conf
+++ after: /var/folders/qn/0rjtdq0d09qbfzv9kflrns2rym3279/T/tmpixfc1M/postgresql-conf.j2
@@ -649,3 +649,4 @@
 
 listen_addresses = '*'
 
+shared_buffers = 8003MB

changed: [test51-database]
```

manics:
Is this going into prod51 too?

PR #131, opened 2018-08-21
Bump all Ansible roles to latest release

manics:
We seem to be lagging behind a bit.

Working on the assumption that we're correctly using semver only the following need to be reviewed:
- [ ] openmicroscopy.cadvisor 0.1.1 → 0.2.0
- [ ] openmicroscopy.omero-web 1.0.0-m3 → 2.0.0
- [ ] openmicroscopy.omero-web-django-prometheus 0.1.0 → 0.2.0

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -10,13 +10,13 @@
   version: 1.0.0
 
 - src: openmicroscopy.anonymous-ftp
-  version: 0.1.0
+  version: 0.1.1
 
 - src: openmicroscopy.basedeps
   version: 1.0.0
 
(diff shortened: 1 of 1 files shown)


sbesson:
Had no issue deploying `test52` with this changes as well as #134. The only breaking change I was unsure about is the one related to the `omero-web`. Another run of the `deploy` phase does not show any change so the configuration of the IDR is probably unaffected.
All the version bumps make sense. Merging this into `master` to be inline with the latest version of the roles. If issues are reported due to the deployment, we can review and address them in follow-up PRs.

PR #132, opened 2018-08-30
Prod51 cleanup

sbesson:
Cleanup of a couple of playbooks reported during the release of IDR 0.5.1

- 5291936 removes `dockermanager-data` from the list of expected volumes
- 8d8530e extends the log archiving playbook to handle `haproxy` and `prometheus`

Looking at the content of `archive-logs.yml` the logic could even be further refactored as a list of tasks applied to 2 groups (`proxy-hosts` and `management-hosts`) with the log/services variables moved to `group_vars`. Since these playbooks are under a subdirectory, this would probably require moving them under the top-level folder and/or symlinking `group_vars`. Happy to stick with the in-playbook `vars` addition for now if this would not be worth it

### DIFF
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -11,52 +11,66 @@
 
 
 - hosts: "{{ idr_environment | default('idr') }}-proxy-hosts"
+  vars:
+    services:
+      - nginx
+      - haproxy
   tasks:
 
-  - name: Archive proxy nginx logs
+  - name: Archive proxy logs
     become: yes
     archive:
-      dest: /tmp/{{ decommission_archive_prefix }}-nginx.tar.gz
+      dest: /tmp/{{ decommission_archive_prefix }}-{{ item }}.tar.gz
       format: gz
-      path: /var/log/nginx
-    register: _decommission_archive_nginx
+      path: /var/log/{{item}}
+    register: _decommission_archive_proxy
+    with_items: "{{ services }}"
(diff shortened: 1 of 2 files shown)


sbesson:
I had some issue with the playbook hanging out during the prometheus logs archiving. While trying to execute the `tar` operation on the node, I realized it was failing with

```
tar: /var/lib/docker/volumes/prometheus-data/_data: file changed as we read it
```

After stopping manually the `prometheus` container, the last version of the playbook successfully downloads all logs. Is it worth adding a preliminary step to stop the `prometheus` container to this playbook?

manics:
There's already a shutdown-services playbook, which stops stops and disables most services including docker: https://github.com/IDR/deployment/blob/IDR-0.5.0/ansible/decommission/archive-instance-services.yml#L24

sbesson:
Thanks @manics. So assuming we update the post-release item to execute `archive-instance-services.yml` followed by `archive-logs.yml`, this should cover everything that had been noticed during the decommissioning of `prod50`. Anything else worth considering here or should we merge?

manics:
The only caveat is since archive-instance-services stops and disables most services it's no longer possible to do a straight swap of the previous production into production, you'd need to re-enable the services, for instance by running the main deployment playbooks. In practice we've never had to do that so I'm fine with it.

manics:
As a todo, we should add these steps to https://github.com/IDR/deployment/tree/master/docs

PR #133, opened 2018-09-03
Install and config for django-cors-headers

will-moore:
See https://trello.com/c/EpmQzvoy/2-json-api-publicly-accessible-with-cors

Trying to add CORS support as described at https://docs.openmicroscopy.org/omero/5.4.7/sysadmins/unix/install-web.html#setting-up-cors

Not sure I quite understand how to perform these append steps:
```
$ bin/omero config append omero.web.middleware '{"index": 0.5, "class": "corsheaders.middleware.CorsMiddleware"}'
$ bin/omero config append omero.web.middleware '{"index": 10, "class": "corsheaders.middleware.CorsPostCsrfMiddleware"}'
```
since all the other configs seem to be "set" but I uncommented out the ```omero_web_apps_config_append``` and added yml there which I hope will be appended as JSON, but I don't really know if/how this works or how to test it.

cc @manics @sbesson

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -163,9 +163,6 @@ omero_web_config_set:
   omero.web.public.cache.timeout: 60
   omero.web.page_size: 500
 
-  # NB: See additional CORS config below
-  omero.web.cors_origin_allow_all: True
-
 ######################################################################
 # Plugins and additional web configuration
 
@@ -264,6 +261,7 @@ omero_web_apps_config_append:
     class: "corsheaders.middleware.CorsPostCsrfMiddleware"
 # set...
 omero_web_apps_config_set:
+  omero.web.cors_origin_allow_all: True
   omero.web.mapr.config:
   - menu: "gene"
     config:
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -170,8 +170,10 @@ omero_web_config_set:
 
 omero_web_apps_packages:
 - "omero-mapr==0.2.3"
+- "django-cors-headers"
 omero_web_apps_names:
 - omero_mapr
+- corsheaders
 
 omero_web_apps_top_links:
 - label: Studies
@@ -251,8 +253,15 @@ omero_web_apps_ui_metadata_panes:
   label: "Others"
   index: 6
 
-# Additional plugin config (append)
-#omero_web_apps_config_append:
+# Additional plugin config
+# append...
+omero_web_apps_config_append:
+  omero.web.middleware:
+  - index: 0.5
+    class": "corsheaders.middleware.CorsMiddleware"
+  - index: 10
+    class": "corsheaders.middleware.CorsPostCsrfMiddleware"
+# set...
 omero_web_apps_config_set:
   omero.web.mapr.config:
   - menu: "gene"
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -258,9 +258,9 @@ omero_web_apps_ui_metadata_panes:
 omero_web_apps_config_append:
   omero.web.middleware:
   - index: 0.5
-    class": "corsheaders.middleware.CorsMiddleware"
+    class: "corsheaders.middleware.CorsMiddleware"
   - index: 10
-    class": "corsheaders.middleware.CorsPostCsrfMiddleware"
+    class: "corsheaders.middleware.CorsPostCsrfMiddleware"
 # set...
 omero_web_apps_config_set:
   omero.web.mapr.config:
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -163,7 +163,8 @@ omero_web_config_set:
   omero.web.public.cache.timeout: 60
   omero.web.page_size: 500
 
-
+  # NB: See additional CORS config below
+  omero.web.cors_origin_allow_all: True
 
 ######################################################################
 # Plugins and additional web configuration
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -94,7 +94,7 @@
   version: 1.0.0-m3
 
 - src: openmicroscopy.omero-web-apps
-  version: 0.1.0
+  version: 0.1.1
 
 - src: openmicroscopy.openstack-volume-storage
   version: 1.1.0


will-moore:
Travis failed with:
```
[0m[0;31mfatal: [idr-omeroreadonly-docker]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service omero-web: Job for omero-web.service failed because the control process exited with error code. See \"systemctl status omero-web.service\" and \"journalctl -xe\" for details.\n"}[0m[0m
```

will-moore:
Seb: "so ```pip install ome-ansible-molecule-dependencies``` should install ansible + molecule
then ```cd ansible; molecule test``` is what Travis executes.
In your case, locally you can probably execute the first phase of the ```test``` i.e. ```cd ansible; molecule converge```
this should create the instance locally and try to provision a minimal IDR
but *not* tear down the containers so you can inspect the error in OMERO.web

will-moore:
```
$ molecule converge
...
- downloading role 'nginx', owned by openmicroscopy
- downloading role from https://github.com/openmicroscopy/ansible-role-nginx/archive/1.0.0.tar.gz
 [ERROR]: failed to download the file: Failed to connect to github.com at port
443: [Errno 8] nodename nor servname provided, or not known

 [WARNING]: - openmicroscopy.nginx was NOT installed successfully.

ERROR! - you can use --ignore-errors to skip failed roles and finish processing the list.
ERROR: 

  RAN: /usr/local/bin/ansible-galaxy install --roles-path=.molecule/roles --force --role-file=requirements.yml

  STDOUT:
  STDERR:
```

will-moore:
@sbesson  I got ```[Errno 8] nodename nor servname provided, or not known```  (see above). Any ideas? Thanks.

sbesson:
@will-moore this is a transient error when downloading roles. Which version of ansible do you have installed? Can you pre-run `/usr/local/bin/ansible-galaxy install --roles-path=.molecule/roles --role-file=requirements.yml`, calling it until all roles are downloaded, so that the roles are installed before executing the playbook?

Generally, I am also facing this issue when creating fresh deployments environments for IDR. I am surprised that Travis is not failing with more timeouts. Is there a known workaround? /cc @manics

ansible/group_vars/omero-hosts.yml
```diff
@@ -251,8 +253,15 @@ omero_web_apps_ui_metadata_panes:
...
   index: 6
 
-# Additional plugin config (append)
-#omero_web_apps_config_append:
+# Additional plugin config
+# append...
+omero_web_apps_config_append:
+  omero.web.middleware:
+  - index: 0.5
+    class": "corsheaders.middleware.CorsMiddleware"
```
manics: Typo: trailing `"` after `class`, same below.

will-moore:
Thanks @manics - Fixed now.

manics:
Testing with molecule locally:
```
[root@390f87fb394f /]# systemctl status omero-web -l
● omero-web.service - OMERO.web
   Loaded: loaded (/etc/systemd/system/omero-web.service; enabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Tue 2018-09-18 09:46:08 UTC; 12min ago
  Process: 31466 ExecStop=/opt/omero/web/venv/bin/python /opt/omero/web/OMERO.web/bin/omero web stop (code=exited, status=0/SUCCESS)
  Process: 31488 ExecStart=/opt/omero/web/venv/bin/python /opt/omero/web/OMERO.web/bin/omero web start (code=exited, status=170)
  Process: 31473 ExecStartPre=/opt/omero/web/config/omero-web-config-update.sh (code=exited, status=0/SUCCESS)
 Main PID: 12942 (code=exited, status=0/SUCCESS)

Sep 18 09:46:04 390f87fb394f systemd[1]: Starting OMERO.web...
Sep 18 09:46:08 390f87fb394f python[31488]: list indices must be integers, not str
Sep 18 09:46:08 390f87fb394f systemd[1]: omero-web.service: control process exited, code=exited status=170
Sep 18 09:46:08 390f87fb394f systemd[1]: Failed to start OMERO.web.
Sep 18 09:46:08 390f87fb394f systemd[1]: Unit omero-web.service entered failed state.
Sep 18 09:46:08 390f87fb394f systemd[1]: omero-web.service failed.
```
```
[root@390f87fb394f /]# grep omero.web.middleware /opt/omero/web/config/omero-web-apps.omero
config append -- omero.web.middleware '[{"index": 0.5, "class": "corsheaders.middleware.CorsMiddleware"}, {"index": 10, "class": "corsheaders.middleware.CorsPostCsrfMiddleware"}]'
```

There's a bug in the omero-web-apps role which means wou can't append multiple values to the same key: https://github.com/openmicroscopy/ansible-role-omero-web-apps/blob/0.1.0/templates/omero-web-apps-omero.j2#L18

manics:
@will-moore The bug should be fixed by https://github.com/openmicroscopy/ansible-role-omero-web-apps/pull/3
If you cherry-pick https://github.com/manics/deployment/commit/93f3335ca56d73995dda4f3f14c43ea62605c453 to this branch travis should pass.

manics:
This change only affects web so it should be safe to deploy whilst the import is in progress.

sbesson:
Understood. If you feel it is safe, feel free to redeploy against `test52` so that we can test this PR.

will-moore:
Let me know when/where this is deployed so I can try to test CORS is working.

will-moore:
To test, you should be able to use https://github.com/openmicroscopy/openmicroscopy/blob/develop/examples/Training/javascript/index.html
```
cd Training/javascript
python -m SimpleHTTPServer
```
Then login with URL of deployed server etc. public / public?

will-moore:
One way to test this is to open any web page with a different domain from the one we're testing, that uses jQuery (e.g. http://idr.openmicroscopy.org/webclient/userdata/) and paste this into the dev-tools console ```$.getJSON("https://idr-testing.openmicroscopy.org/webclient/api/containers/")```.

This works for my local server (where I have configured CORS) 
```$.getJSON("http://localhost:4080/webclient/api/containers/")```
and I can see in the Network tab that the Response Header has ```Access-Control-Allow-Origin``` set to the domain I'm calling from:

screen shot 2018-09-19 at 12 52 27

However, with ```idr-testing.openmicroscopy.org``` this is missing so we can't access the response:

screen shot 2018-09-19 at 12 51 07

File: ansible/group_vars/omero-hosts.yml [hunk 2]
```diff
@@ -163,15 +163,18 @@ omero_web_config_set:
   omero.web.public.cache.timeout: 60
   omero.web.page_size: 500
 
-
+  # NB: See additional CORS config below
+  omero.web.cors_origin_allow_all: True
```
manics: Minor point, but since the rest of the CORS config is in `omero_web_apps_*` I think this should go under [`omero_web_apps_config_set:`](https://github.com/IDR/deployment/pull/133/files#diff-564d3b63a1550c7a5968bd56c4adb048R266) instead

will-moore: Done

manics:
I've deployed this change and the CORS headers still aren't showing, even after tunnelling through the proxy and connecting directly to the back-end.

manics:
I've opened up public access to `/api` https://github.com/IDR/deployment/pull/136

```
$ curl --dump-header - -H "Origin: http://example.com" https://test52/api/v0/
HTTP/1.1 200 OK
Server: nginx/1.14.0
Date: Fri, 21 Sep 2018 13:14:43 GMT
Content-Type: application/json
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Origin
X-OMERO-ApiVersion: 0.1
Access-Control-Allow-Origin: http://example.com
Access-Control-Allow-Credentials: true

...
```

will-moore:
This is also working for me now, testing in the browser and curl against test52.

sbesson:
Thanks both for driving this. So I understand #136 is required in order to work as it does right now on `test52`, is that correct?. Do we foresee a need for caching the API URLs or is this effetively the same as queries that would be done via the API in the notebook? Does anything need to happen to make sure we monitor the usage/response of these URLs?

will-moore:
@sbesson I don't think that #136 should be necessary for the behaviour in https://github.com/IDR/deployment/pull/133#issuecomment-423526952 with the base /api/ URL e.g. "http://idr.openmicroscopy.org/api/" since this base URL doesn't require login (or public user) whereas e.g. "http://idr.openmicroscopy.org/api/v0/m/screens/" does (currently see ```"message": "Not logged in"```).
However, public access to those other URLs is certainly very useful so it would be great to have both these PRs merged.
All of the JSON API calls map to stateless services and the queries are limited in terms of joins and max number of objects, so caching shouldn't be so critical.
Thanks.

sbesson:
--include

will-moore:
Hmmm - Something strange here. This is working for ```/api/``` URLS but not for ```/webclient/api/``` URLS.
E.g. this works
```
$.getJSON("https://idr.openmicroscopy.org/api/v0/m/projects/")
```
but this doesn't
```
$.getJSON("https://idr.openmicroscopy.org/webclient/api/annotations/?type=map&image=1884807")
```
Fails with:
Access to XMLHttpRequest at 'https://idr.openmicroscopy.org/webclient/api/annotations/?type=map&image=1884807&_=1542892980171' from origin 'http://localhost:4080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

Any ideas what's up?
(playing with a side project: https://gitlab.com/openmicroscopy/incubator/omero-idr-gallery)

manics:
I can't see anything obvious, does it work for other servers?

manics:
It's an nginx caching problem, https://idr.openmicroscopy.org/webclient/api/annotations/?type=map&image=1884807 is already cached along with the headers

manics:
E.g. if you query one of the new datasets the response includes the `access-control-*` headers as expected:
```
$ curl -s --dump-header - -H "Origin: http://blah" 'https://idr.openmicroscopy.org/webclient/api/annotations/?type=map&image=397
7543'
HTTP/2 200
server: nginx/1.14.0
date: Thu, 22 Nov 2018 15:13:48 GMT
content-type: application/json
vary: Cookie, Origin
access-control-allow-origin: http://blah
access-control-allow-credentials: true

...
```

will-moore:
OK, thanks. That'll be fine for now. I guess those older URLs will get un-cached at some point?

manics:
Yes, we'll have to wipe the entire cache at some point.

PR #134, opened 2018-09-04
test52: OMERO.server changes

sbesson:
This PR contains OMERO.server changes for `test52`:

- 8895f22 bumps the version of OMERO to use `0.5.1-rc1` which includes  the 5.4.8-rc1 tag of OMERO with all the parallel import features  - see https://trello.com/c/i0JAPzgB/13-omero-548-rc1
-  1452eed and 164c846 increase the height and the width of the maximum plane size to import the first run of HPA data - see https://trello.com/c/gBAAQTIY/47-hpa-run-1-idr0043
- 4c4c25b updates `omero.fs.repo.path` to contend with parallel imports using the `-%hash` suffix

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -57,7 +57,7 @@ omero_server_omego_additional_args: "{{ idr_omero_omego_additional_args }}"
 omero_server_config_set:
   omero.db.poolsize: 25
   omero.jvmcfg.heap_size.blitz: "24G"
-  omero.fs.repo.path: "%user%_%userId%_%thread%//%year%-%month%/%day%/%time%"
+  omero.fs.repo.path: "%user%_%userId%/%thread%//%year%-%month%/%day%/%time%"
   # public user doesn;t share omero session between visitors
   #omero.sessions.timeout: 3600000
   omero.policy.binary_access: "+read,+write,-image,-plate"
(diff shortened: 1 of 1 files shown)


joshmoore:
http://downloads.openmicroscopy.org/idr/omero/0.5.1-rc1/ :+1: LGTM

joshmoore:
(Though of course I'd say that... :smile:)

joshmoore:
...still looks ok.

sbesson:
The last set of commits:

- pushed away the max plane changes to #137 as we have not agreed on the run to be loaded into `prod52` yet. This should allow this PR to get merged indepenently
- bumped the `idr_omero_release` to 0.5.1 containing the changes from the OMERO 5.4.8-ga3 /cc @joshmoore @jburel 
- updated `omero.fs.repo.path` to use `%thread%` instead of `%hash%` earlier in the directory path. Although less attractive, this should prevent PSQL race conditions reported right after midnight during the latest long-running `hpa_run_01` import /cc @mtbc @dominikl

ansible/group_vars/omero-hosts.yml
```diff
@@ -57,6 +57,7 @@ omero_server_omego_additional_args: "{{ idr_omero_omego_additional_args }}"
 omero_server_config_set:
   omero.db.poolsize: 25
   omero.jvmcfg.heap_size.blitz: "24G"
+  omero.fs.repo.path: "%user%_%userId%/%thread%//%year%-%month%/%day%/%time%"
```
mtbc: :+1:

dominikl:
Just curious, so even with the `%hash%` option there was a collision with the path name?

mtbc:
If `%hash%` comes after `%day%` there's a problem at midnight when multiple threads may try to create the directory for the new day. `%hash%` would (very?) *probably* work at a higher directory level but `%thread%` is sure to.

PR #137, opened 2018-09-21
Adjust maximum plane height and width

sbesson:
This change is required to import the first run of Human Protein Atlas data without generating TBs of OMERO pyramids.


sbesson:
Closing as the first batches of 1000 HPA antibodies published in IDR will be of size 3Kx3K. For the larger images, alternate solutions will be investigated including the generation of pyramidal OME-TIFF.

PR #139, opened 2018-09-26
Bump to IDR OMERO 0.5.2

sbesson:
The only difference with b8e4b155ad4fcca0027fd1ce6a4bb9faa50c189f are the last set of PRs included in the public release of OMERO 5.4.8 (see https://github.com/openmicroscopy/openmicroscopy/pull/5872). This should not have any impact in the case of IDR but reduces the delta between the IDR version of OMERO and mainline OMERO to its current minimum.


sbesson:
Now deployed on `test52`

PR #140, opened 2018-10-15
Auto-restart after SSL update, revert to previous Nginx logging format

manics:
- [x] --depends-on https://github.com/openmicroscopy/ansible-role-ssl-certificate/pull/11

This also reverts `openmicroscopy.nginx-proxy` role to `1.7.0` to ensure the Nginx log format is identical to the existing production (prod51) format. Longer term we either need to modify the role to a support the IDR's log format, or update all our log analytics tools to accept both formats.


sbesson:
@manics was this deployed against `test52`/`prod52`? Is there some verification to be done re the NGINX log format?

PR #143, opened 2018-11-07
Fix variable errors in archive logs msg

manics:
These errors were discovered whilst running the playbook today on prod51: https://trello.com/c/Jww9YsQ0/12-archive-logs


joshmoore:
Does this mean something was overwritten?

PR #144, opened 2018-11-08
Ansible playbook to dump postgres DB

manics:
Another decommissioning playbook to take a dump of the database. Requires Ansible 2.4

Used in https://trello.com/c/oEOqeKgR/31-re-snapshot-db

### DIFF
--- a/ansible/decommission/archive-db.yml
+++ b/ansible/decommission/archive-db.yml
@@ -1,0 +1,59 @@
+# Dump and fetch a database
+
+- hosts: "{{ idr_environment | default('idr') }}-database-hosts"
+  tasks:
+
+  - name: Get archive prefix
+    set_fact:
+      decommission_archive_prefix: "{{ ansible_hostname }}-{{ ansible_date_time.date | replace('-', '') }}"
+
+  # postgresql_db dump is not idempotent
+  - name: Check if dump exists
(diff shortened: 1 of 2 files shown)


manics:
Only the directory format supports parallel exports. I think that could be passed in `target_opts` but it means an extra step to zip up the directories. Since the dump is fairly small (2.9GB) I think it's easier to keep it as a single dump file.

If we're completely removing the dump script we'll need to ensure the database dump is still present for rsync access.

sbesson:
The cards have not been updated for `prod60` which means I used the old script. Should I use this instead? What is the preferential way to allow the `rsync`?

manics:
You'd probably need to run the dump task on omeroreadwrite. Can look at that in a future PR, but for now just use this for the backup dump?

sbesson:
I have been trying to execute this playbook in the context of `prod52` decommissioning. A single 3.3GB `pgdump` was created under `/tmp` as expected however the playbook seemed to be hanging at the `Dump database` step. I will retry tomorrow but is there anything obvious I might be missing.

manics:
Sounds about right, it takes a while with no apparent output.

sbesson:
Ah yes, it completed. Just need to be a bit more patient.

PR #147, opened 2018-11-19
0.6.0 rc3

sbesson:
Includes the changes merged in https://github.com/openmicroscopy/openmicroscopy/pull/5909 and to be released as OMERO 5.4.10 /cc @jburel 

The  `omero.web.user_dropdown` option should remove the login dropdown from the read-only deployments (proxied by the `idr-proxy`) but maintain an Login option on the read-write deployment for edition of the top-level containers.

Since `test60` is currently used for testing the Bio-Formats cache regeneration, proposing to deploy this on `prod60`


sbesson:
Successfully deployed on `prod60`. See the read-only Web deployment

screen shot 2018-11-21 at 09 14 45

and the readwrite Web deployment 

screen shot 2018-11-21 at 09 14 47

PR #150, opened 2018-12-03
Add dataset and well to iviewer

manics:
Based on latest config in https://github.com/ome/omero-iviewer/blob/v0.6.0/plugin/omero_iviewer/README.rst


sbesson:
Deployed on `test61`

PR #152, opened 2018-12-03
Add overarching decommissioning playbook

sbesson:
See https://github.com/IDR/deployment/pull/144#pullrequestreview-173398053

While decommissioning `prod52`, I realized that the steps had to be run in a specific order. This playbook captures this order to simplify.

Note the steps aggregating/copying the various dumps/archives to `/uod/idr/versions` is not implemented but could probably be conceived as a folllow-up playbook.

### DIFF
--- a/ansible/decommission/playbook.yml
+++ b/ansible/decommission/playbook.yml
@@ -1,0 +1,4 @@
+---
+- include: archive-db.yml
+- include: archive-instance-services.yml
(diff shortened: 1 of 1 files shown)


sbesson:
No special options at the moment but I propose to start consuming and expanding this playbook as the single reference one for decommissioning purposes.

PR #153, opened 2018-12-04
Molecule test for FTP server

manics:
Adds a new Molecule scenario for testing idr-ftp.


.travis.yml
```diff
@@ -18,4 +18,8 @@ script:
     --exclude ../kubespray/
     --exclude idr-kubernetes-apply.yml
     --exclude idr-reset-users-groups.yml
```
manics: Obviously it would make sense to move the above common tasks out of the matrix build instead of running them twice, but it's not clear how from reading the Travis docs, and given the time taken by each re-run I'm not going to attempt it unless someone has a known working example.

sbesson: I do not have any production example either in any of our repos. That being said, I am very interested in converging towards that as we will see the need in other repos. My initial idea would be to try something along the lines of https://docs.travis-ci.com/user/build-stages#build-stages-and-build-matrix-expansion maybe using `molecule syntax` first to prototype and then switch to `molecule test` when the config is working?

manics: Thinking about it it might be possible to have a `syntax-only` molecule scenario in which case this the setup here works nicely. Mind if that's in later PR?

sbesson: Looking at the details of the [Travis completion time](https://travis-ci.org/IDR/deployment/builds/463310820)

|  Scenario |  Total time |  Molecule test time |
|-----------|-------------|---------------------|
| publicidr | 24min 46s   | 23min 12s           |
| ftp       |  6min 3s    | 4min36s             |

Previous executions were ~24-25min so I do not think this gets us to a much worse situation since you are effectively testing a completely separate deployment path.

.travis.yml
```diff
@@ -18,4 +18,8 @@ script:
...
-  - cd ansible; molecule test
+  - cd ansible; molecule test -s $SCENARIO
+
+env:
+  - SCENARIO=publicidr
```
sbesson: It's been a while we did not come to a naming discussion. I am still very unconvinced by the usage of `publicidr`, minimally because e.g. `idr-analysis` is similarly public. If we are considering any of these are components of the overarching IDR deployment, an alternate proposal for the types would be something of  `ingress/datasets/analysis` as a way to represent the aim of each infrastructure.

manics: I thought this might come up 😂. I'm just using the names from https://github.com/IDR/deployment/blob/IDR-0.6.0/ansible/openstack-create-infrastructure.yml

joshmoore: :+1: for changing. We have enough external confusion with "IDR" anyway.

sbesson: Indeed 😄 . And at the time the split was introduced, we didn't manage to get to a better naming consensus but knew we might eventually come back to it - see https://github.com/IDR/deployment/pull/95#pullrequestreview-106083611. Happy to consider other alternatives obviously.

manics: How about get this PR in, then whoever's brave enough to make a proposal can open a renaming PR?

sbesson: I am up for have this integrated together with #155 so that the new addition is tested by `molecule`. Regarding the new variable name

https://github.com/IDR/deployment/blob/94876d81a84b1ed48515ff5e163fbb0cde5762bb/ansible/openstack-create-infrastructure.yml#L4-L5

Is there an easy Jinja/Ansible way to check 2 variables (i.e. maintain backwards compatibility)? Or is the only way to have some kind of hard break?

manics: We could, but since we're fully in control of this repository I think adding logic to handle alternative variable names makes maintenance more complex for no real benefit.

PR #154, opened 2018-12-04
Set omero.web.api.max_limit 1000

manics:
This should allow up to 1000 ROIs to be shown in iviewer, but will also allow larger API queries. Needs testing.

See https://trello.com/c/LdGRTtZI/1-omeroiviewer-as-the-default-viewer#comment-5bf52300aafbb91554148bfe


sbesson:
Deployed on `test61`

jburel:
Need to see how the UI will handle that and check the other queries

PR #155, opened 2018-12-13
idrftp alternate user account

manics:
Creates a locked local account on the IDR FTP server.
This account can be manually unlocked and a password set when required.


manics:
Updated to `ftp-upload`

joshmoore:
Green. Merging. We can always review behavior again later.

PR #156, opened 2018-12-17
Bump DB volume size to 300

sbesson:
Candidate for `prod62`. As demonstrated by https://github.com/IDR/idr.openmicroscopy.org/pull/40, we are getting closer to the volume limit on the production IDR database instance. This raises the size to 300GB instead.

### DIFF
--- a/ansible/openstack-create-publicidr.yml
+++ b/ansible/openstack-create-publicidr.yml
@@ -101,7 +101,7 @@
     # IDR Volumes
 
     - role: openmicroscopy.openstack-volume-storage
-      openstack_volume_size: 250
+      openstack_volume_size: 300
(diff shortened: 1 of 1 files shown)


manics:
screen shot 2019-01-28 at 11 13 05

Given the size of the HPA datasets I think 500GB would be better

sbesson:
Pushed. Note that together with #163, this effectively means an volume size increase of 750GB per IDR stack. From the EBI tenancy, this is within the range of our allowance and matches the ongoing dataset import plan.

manics:
This was manually deployed to prod63: https://trello.com/c/6AFjJxvJ/4-resize-database-volume

PR #157, opened 2018-12-20
k8s API access must be tunnelled via k8sproxy

manics:
Prevent external K8s API access.

This increases security by requiring K8s API access to be either tunnelled via the proxy (ssh proxy, connect to localhost), or from within the cluster (the GitLab runner is hosted inside the cluster and therefore has internal API access).

Relevant playbook: `ansible/k8s-k8sproxy.yml`


manics:
This has been deployed. Private credentials have been updated to assume ssh localhost tunnel.

joshmoore:
Better safe than sorry. Can always review when/if someone needs non-localhost access.

PR #162, opened 2019-01-16
Update website and stats to 2019.01.16

manics:
prod62 release


manics:
This is deployed on prod62

Issue #141, opened 2018-10-15
Upgrade to nginx-proxy 1.8.0 ansible role breaks the log format


manics:
See https://github.com/IDR/deployment/pull/140

We either need to:
- modify the role to a support the IDR's log format
- update all our IDR log analytics tools to accept both formats

PR #165, opened 2019-02-13
Change cache_validity to 1d, background update, update nginx-proxy

manics:
This only decreases the Nginx expiry cache time, the inactive time is unchanged. This means if a cache update fails Nginx should return the stale version instead of failing (and should continue to refresh in the background)

Requires:
- [x] https://github.com/openmicroscopy/ansible-role-nginx-proxy/pull/21

Closes https://github.com/IDR/deployment/issues/141

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -20,7 +20,12 @@ nginx_proxy_http2: True
 ######################################################################
 # OMERO.web proxy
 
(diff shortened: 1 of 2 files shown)


manics:
Commit https://github.com/IDR/deployment/pull/165/commits/2745108f6272c1e9e60f480863d5d802b3be5374 is deployed to `test64`

manics:
Commit 9a075859faa00fd65e7bc4cce320b41e4abe494e is deployed to `test64`

PR #168, opened 2019-03-01
Upgrade to OMERO 5.5 and Bio-Formats 6 

sbesson:
## Context

The goal of this PR is to upgrade the sofware deployed on the Image Data Resource to use the code for the latest major versions of OMERO 5.5 and Bio-Formats 6.

Up to IDR 0.5.0, [IDR](https://idr.openmicroscopy.org/) was deployed used a custom build of OMERO including the first prototype of the OMERO read-only functionality as well as a custom build of Bio-Formats with changes written specifically for IDR studies (ScreenReader, FlexReader...).

A large portion of the OMERO 5.4.0 work included the backporting of the IDR features into the mainline OMERO distributed by OME.  With further backporting, the changes between an early milestone of OMERO 5.5.1 and the build using in IDR 0.6.x is mainly limited the custom version of IDR Bio-Formats and the disabled OMERO scripts - see  https://github.com/openmicroscopy/openmicroscopy/compare/v5.5.0-m1...IDR:IDR-0.6.0.

A new version of the IDR Bio-Formats library merging Bio-Formats 6.2.0 with the custom modifications still required for the historical IDR studies has been tested and relased as 0.6.0.

## Changes

This PR makes the necessary changes to deploy the upgraded OMERO/Bio-Formats stack into IDR. The following deployment changes are made:

- the deployment now uses the upstream OMERO version rather than a custom IDR build and the `idr_omero_version` variable is set to 5.5.1
- the `idr_omero_upgrade` and `idr_omero_ice_version` variables are removed as the playbook use the default role values (`True` and `3.6`)
- additional tasks are added to the OMERO read-write and read-only playbooks to customize the server deployment by overriding the Bio-Formats library JARs under `lib/server` and `lib/client` and deleting the scripts directory.

## Testing

This should be deployed on `test70` and functional testing should ensure that:

- the first plate/dataset of each study is readable following the recaching of the memo files
- the  OMERO scripts are disabled


sbesson:
Deployed on `test70`

ansible/group_vars/omero-hosts.yml
```diff
@@ -6,15 +6,18 @@
...
 
-idr_omero_release: "0.6.1"
+idr_omero_release: "5.5.1"
 idr_omero_ice_version: "3.6"
-# Upgrades are normally disabled, set to True upgrade OMERO:
-idr_omero_upgrade: False
-# The IDR version system doesn't match the OMERO.server version
-omero_server_checkupgrade_comparator: '!='
-omero_web_checkupgrade_comparator: '!='
+idr_omero_upgrade: True
```
manics: The default in the omero server and web roles is to upgrade if the version doesn't match, so it should be safe to delete these variables completely:
```
$ git grep idr_omero_upgrade FETCH_HEAD
FETCH_HEAD:ansible/group_vars/omero-hosts.yml:idr_omero_upgrade: True
FETCH_HEAD:ansible/group_vars/omero-hosts.yml:omero_server_upgrade: "{{ idr_omero_upgrade }}"
FETCH_HEAD:ansible/group_vars/omero-hosts.yml:omero_web_upgrade: "{{ idr_omero_upgrade }}"
FETCH_HEAD:ansible/var-overrides/omero-ci-build.yml:idr_omero_upgrade: True
```

ansible/idr-omero-readonly.yml
```diff
@@ -185,4 +185,23 @@
   roles:
   - role: openmicroscopy.omero-server
 
+  tasks:
+    - name: Override Bio-Formats artifacts in lib/server
```
manics: Minor nitpick, can you add an empty line between multiple tasks in this and other playbooks?

ansible/idr-omero.yml
```diff
@@ -55,4 +55,31 @@
...
+    - name: Override Bio-Formats artifacts in lib/server
+      become: yes
+      become_user: "{{ omero_server_system_user }}"
+      get_url:
+        url: "{{ idr_bf_baseurl }}/{{ item }}/{{ idr_bf_release }}/{{ item }}-{{ idr_bf_release }}.jar"
+        dest: "{{ omero_server_basedir }}/{{ omero_server_symlink }}/lib/server/{{ item }}.jar"
+        force: yes
+      with_items: "{{ idr_bf_components }}"
+      notify: omero-server restart omero-server
+    - name: Override Bio-Formats artifacts in lib/client
```
manics: `Override Bio-Formats artifacts in lib/client` is missing from `idr-omero-readonly.yml`.

If you're concerned about duplicate code one option is to move it into another playbook that's called after: https://github.com/IDR/deployment/blob/IDR-0.6.8-1/ansible/idr-01-install-idr.yml#L15

Something like
```yaml
roles:
  - ome.omero_common
tasks:
  - name: Override Bio-Formats artifacts in lib/server
    notify: restart omero-server # Generic restart handler from ome.omero_common instead of the omero-server internal one
  - name: Override Bio-Formats artifacts in lib/client
    notify: restart omero-server
  - name: Remove OMERO scripts
    notify: restart omero-server
```

ansible/group_vars/omero-hosts.yml
sbesson: See 1f3326b. I also remove the Ice variables which were redefining the role default

ansible/idr-omero-readonly.yml
sbesson: Done in the separate playbook

ansible/idr-omero.yml
sbesson: Thanks for the suggestion. I initially skipped `lib/client` from `idr-omero-readonly.yml` as we would not run a `bin/omero import` in these servers. I like the idea of migrating the custom IDR step into a separate playbook, both for compartmentalization and ease of testing of future IDR/Bio-Formats versions. Pushed in 3678c87

sbesson: Ah, from https://travis-ci.org/IDR/deployment/jobs/565055434, the downside of using `omero_common` is the loss of knowledge about the OMERO.server system user

````
    fatal: [idr-omeroreadonly-docker]: FAILED! => {"msg": "The field 'become_user' has an invalid value, which includes an undefined variable. The error was: 'omero_server_system_user' is undefined"}
```

manics: Get rid of `become_user`? It should be fine to run the copy as `root`, there's no reason it needs to be writeable by OMERO (long term it'd be nice if most of OMERO.server was owned by root but that's a separate discussion).

manics: I think it's also fine to hard-code the server path `/opt/omero/server/OMERO.server` or `{{ omero_common_basedir }}/server/OMERO.server` instead of depending on the omero-server role variables, e.g. as in https://github.com/IDR/deployment/blob/IDR-0.6.8-1/ansible/management-prometheus.yml#L46

PR #173, opened 2019-03-28
Cap molecule to <2.20

manics:
A recent update to molecule has led to failures in this repo (might just be linting/formatting). Since other roles such as https://travis-ci.org/openmicroscopy/ansible-role-python-pydata seem fine this caps molecule for this repo only until we have time to investigate and update.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ services:
   - docker
 
 install:
-  - pip install "ome-ansible-molecule<0.5"
+  - pip install "ome-ansible-molecule<0.5" "molecule<2.20"
 
 script:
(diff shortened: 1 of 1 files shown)


sbesson:
I updated and triggered https://travis-ci.org/ome/ansible-roles/jobs/513980095 to assess the impact on existing roles converted to Molecule 2. Overall it seems that the main changes are syntatic (also enforcing the deprecation of `categories` in favor of `galaxy_tags`).

If we are capping, should we do it in the upstream `ome-ansible-molecule` meta-package while we review/roll out the changes?

manics:
Since other roles are affected that makes sense: https://github.com/openmicroscopy/ome-ansible-molecule/pull/15

PR #179, opened 2019-05-22
Bump mapr to 0.3.0a1, install gallery

will-moore:
To allow testing of https://github.com/ome/omero-mapr/pull/46 with https://github.com/ome/omero-gallery/pull/25

PR #180, opened 2019-06-05
Gallery is the idr root

manics:
~/ ➜ /gallery
/cell ➜ /gallery/cell
/tissue ➜ /gallery/tissue~

`/` is now the gallery app

### DIFF
--- a/ansible/files/patches/pr6046-omeroweb-urls.patch
+++ b/ansible/files/patches/pr6046-omeroweb-urls.patch
@@ -1,0 +1,21 @@
+diff --git a/components/tools/OmeroWeb/omeroweb/urls.py b/components/tools/OmeroWeb/omeroweb/urls.py
+index 8e10516e918..cad9195b8c2 100755
+--- a/components/tools/OmeroWeb/omeroweb/urls.py
++++ b/components/tools/OmeroWeb/omeroweb/urls.py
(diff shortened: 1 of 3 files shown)


manics:
Deployed to test64. Travis will fail until gallery is updated with https://github.com/ome/omero-gallery/pull/33 since the molecule proxy test checks for the IDR title string.

joshmoore:
:+1: slash endings are missing for the likes of https://X/cell/

Was hiding the `/gallery/` element from the URL doable?

manics:
Added `/` handling.

Rewrites are not supported by the current Ansible role. In any case Django apps don't support rewriting, they need to know their full location. It might be possible to develop a custom set of rewrite rules but they'd be liable to break when the app is modified.

joshmoore:
> but they'd be liable to break when the app is modified.

With the flip-side being the URLs are guaranteed to break if we drop `/gallery/` moving forward.

manics:
Redirecting is easy, rewriting to hide a URL isn't.

manics:
Deployed to test64

manics:
I can't fix travis until https://github.com/IDR/deployment/pull/179 and https://github.com/will-moore/deployment/pull/1 are merged, since the molecule test  checks the response from `/`

PR #182, opened 2019-06-10
OMERO.web public: change `/gallery` to `/gallery-api`

manics:
I have no idea why this is needed. `/cell` and `/tissue` worked whereas `/gallery-api` didn't. I have no idea why `/gallery` does not seem to be required.

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -104,7 +104,7 @@ idr_omero_web_public_url_filters:
 - webadmin/myphoto/
 - mapr/
 - iviewer/
-- gallery/
+- gallery-api/
 - webclient/(?!({{ idr_omero_web_public_url_filters_webclient_exclude | join('|') }}))
 - webgateway/(?!(archived_files|download_as))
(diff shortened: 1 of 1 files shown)


manics:
Deployed to test64

sbesson:
With limited knowledge, my best supposition is that the only gallery URL that needs to te be added to  the public filter whitelist is the one pertaining to thumbnails. Probably prior to moving the `gallery/` application to the root of IDR in https://github.com/IDR/deployment/pull/180/files#diff-ef7c8698894286449b6719f3e8a72ee6R12, it might have been sufficient to have `gallery/gallery-api`.

In any cases, thumbnails are now accessible. Performance can be tested next with #184. Merging for the joint deployment on `test64` and `prod67`

manics:
@sbesson Will explained in https://github.com/ome/omero-gallery/pull/37#issuecomment-500415646
`gallery-api` is the only URL that connects to omero:4064. Previously the endpoint was called `/api`, which was already whitelisted (and may have conflicted) with the existing OMERO.web `/api`.

PR #184, opened 2019-06-10
Cache /gallery-api/

manics:
This might improve performance slightly

Not deployed, since it makes testing changes on test64 harder


manics:
Deployed on test64 earlier today, and prod67 just now https://openmicroscopy.slack.com/archives/C0K5WAD8A/p1560268244220100

joshmoore:
The changes seem fine and certainly the performance was _much_ faster on all subsequent calls. :+1: After release, we probably should (periodically?) have a system that _isn't_ using caching so we can figure out how to optimize the calls themselves. cc: @will-moore

manics:
Hopefully web-dev-merge can be that system!

joshmoore:
:+1:

PR #185, opened 2019-06-14
Update the location of the idr-ci Jenkins server

sbesson:
Tested during the redeployment of `test70`


sbesson:
Proposing to merge as this should not affect the stable `master` branch but is required to deploy merge OMERO 5.5 servers onto `test70`. @manics alternatively would you rather get this into the new integration branch?

manics:
It's fine, it's not used automatically (and it's probably broken in the current state anyway).

PR #191, opened 2019-07-02
Postgres version

manics:
Pin Postgresql to 9.6.13 in molecule test to workaround https://forum.image.sc/t/fyi-not-possible-to-create-new-omero-instance-with-latest-postgresql-versions/26929

### DIFF
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -100,8 +100,8 @@
   version: 1.1.0
 
 - name: openmicroscopy.postgresql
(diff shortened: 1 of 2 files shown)


ansible/requirements.yml
```diff
@@ -99,8 +99,9 @@
 - src: openmicroscopy.openstack-volume-storage
   version: 1.1.0
 
-- src: openmicroscopy.postgresql
-  version: 3.0.1
+- name: openmicroscopy.postgresql
```
manics: To avoid cascading dependency bumps the new role `src: ome.postgresql` should keep the same `name` for now

PR #193, opened 2019-07-05
Update prod68-proxy

joshmoore:
already deployed with:

```
./playbooks/deploy-idr.sh prod68 expert deployment/ansible/idr-proxy-about.yml --diff
```


joshmoore:
Travis failure is due to pg 9.6.14 issue.

PR #196, opened 2019-07-29
Point kubespray submodule to OME fork

sbesson:


sbesson:
Travis is green. Merging to remove the dependency on the upstream as a submodule. A next action might be to rearchitect the VAE infrastructure and remove the kubespray module.

PR #197, opened 2019-08-01
Rename idr-bastion as idr-hosts and expand the scope of hosts_populate

sbesson:
The primary intent of this PR is to allow any host in the bastion to use the alias

Since `screen` and `parallel` are installed on `omeroreadwrite`, this should be the only change to be able to run https://github.com/IDR/deployment/blob/master/docs/operating-procedures.md#bio-formats-cache-regeneration from the `omeroreadwrite` node (as opposed to the `proxy` node previously

Also unifies all the roles that should be run against all hosts (`hosts_populate`, `sudoers`, `local_accounts`) into a single playbook


manics:
Do you want to use this rename as an opportunity to move the content of https://github.com/IDR/deployment/blob/IDR-0.6.8-1/ansible/idr-local-users.yml to the start of this playbook?

sbesson:
The last commit has been provisionally deployed on `test70`. I had to skip the `ome.hosts_populate` in Travis due to issue with modifying `/etc/hosts` in running Docker containers.

PR #200, opened 2019-08-05
Bump IDR Bio-Formats release to 0.6.1

sbesson:
This should included the fix for reading screenD of the idr0011 study


sbesson:
Now deployed on `test70`. Memo files for `idr0011-ledesmafernandez-dad4/screenD` were correctly regenerated and a smoke testing did not reveal any issue.

Will list for a final team review tomorrow in order to freeze the sofware for `prod70`

PR #201, opened 2019-08-06
inventories: py3 compatibility

manics:
Python 3 fixes for inventories/openstack-idr.py

WARNING: the openstack library does some hidden caching, it's not enough to delete the cache files. If you immediately switch from Python 2 to Python 3 the inventory won't work properly- the SSH proxy autodetection fails. Delete any cache files you can find and wait a bit.

### DIFF
--- a/inventories/openstack-idr.py
+++ b/inventories/openstack-idr.py
@@ -153,7 +153,7 @@ def append_hostvars(hostvars, groups, key, server, namegroup=False):
     except KeyError:
         pass
 
-    for network, ports in server['addresses'].iteritems():
+    for network, ports in server['addresses'].items():
         if ansible_ssh_host:
             break
         for port in ports:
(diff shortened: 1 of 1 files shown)


mtbc:
The Travis failure does not look as if this PR would cause it.

joshmoore:
> The Travis failure does not look as if this PR would cause it.

```
    TASK [ome.omero_web : omero web | get latest downloads url] ********************
    fatal: [idr-omeroreadonly-docker]: FAILED! => {"changed": false, "msg": "Failed to connect to downloads.openmicroscopy.org at port 443: [Errno 110] Connection timed out"}
    fatal: [idr-omeroreadwrite-docker]: FAILED! => {"changed": false, "msg": "Failed to connect to downloads.openmicroscopy.org at port 443: [Errno 110] Connection timed out"}
```

Indeed not. Restarted.

mtbc:
... and passed! :cake:

PR #208, opened 2019-09-23
Bump omero-mapr version to 0.3.2

sbesson:
Includes the two following bug fixes

- https://github.com/ome/omero-mapr/pull/50
- https://github.com/ome/omero-mapr/pull/44

Deployed on `test71`. In addition to executing the playbook, I had to wipe the `omerostatic` cache and reload `nginx`. I wonder whether we want to do that more systematically in the deployment playbooks

```
[sbesson@test71-proxy ~]$ sudo rm -rf /var/cache/nginx/omerostatic/
[sbesson@test71-proxy ~]$ sudo service nginx reload
Redirecting to /bin/systemctl reload nginx.service
```

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -169,7 +169,7 @@ omero_web_config_set:
 # Plugins and additional web configuration
 
 omero_web_apps_packages:
-- omero-mapr==0.3.0
+- omero-mapr==0.3.2
 - omero-iviewer==0.8.1
 - omero-gallery==3.2.0a8
 omero_web_apps_names:


jburel:
Ddi you have to clean omerostatic to access the new version of the plugin?
if this is the case, this should probably be added to the playbook

jburel:
I wonder if we should have the ability in the UI to know which version of the app is installed similar to iviewer or figure

jburel:
https://idr-next.openmicroscopy.org/cell/search/?query=mapr_gene:a5 

502 Bad Gateway
nginx/1.16.1

sbesson:
Yes as mentioned in the description, cleaning the cache was necessary to refresh something like http://idr.openmicroscopy.org/static/mapr/javascript/ome.mapr.ui.autocomplete.js?_0.3 since the JS was updated within patch releases. Adding the `omerostatic` cleanup to the deployment playbook would be an option.

As discussed yesterday, updating the code so that the URL includes the full version e.g. http://idr.openmicroscopy.org/static/mapr/javascript/ome.mapr.ui.autocomplete.js?_0.3.2 would be another.

sbesson:
Deployed on `prod72`

PR #212, opened 2019-11-20
Add study-specific redirects to the download page

sbesson:
This study is scheduled for publication in an upcoming IDR release. As the publication went online, the raw data is available via Aspera. In the interim, this allows the link to redirect to the download page.

Deployed on `test71`. Proposing to deploy on `prod71` immediately and tag 
cf25e5a  as `IDR-0.7.1-1`. The commit will need reverting for `prod72` release.

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -278,6 +278,10 @@ _nginx_proxy_sites:
     # Only add this CORS header to the default public site
     nginx_proxy_additional_directives:
     - "add_header Access-Control-Allow-Origin $allow_origin"
+    # Study redirects
+    - "if ($request_uri ~ /search/\\?query=Name:idr0065) {
+       return 302 /about/download.html;}"
+
 
   # This is a duplicate of the main OMERO.web proxy configuration, but with
   # cache-busting:


ansible/group_vars/proxy-hosts.yml
```diff
@@ -278,6 +278,10 @@ _nginx_proxy_sites:
     # Only add this CORS header to the default public site
     nginx_proxy_additional_directives:
     - "add_header Access-Control-Allow-Origin $allow_origin"
+    # Study redirects
+    - "if ($request_uri ~ /search/\\?query=Name:idr0065) {
+       return 302 /about/download.html;}"
+
```
manics: We already have a section for custom study redirects:
https://github.com/IDR/deployment/blob/97b70b98c64bcdc5e3bcac5ffb206e59856cb208/ansible/group_vars/proxy-hosts.yml#L206-L228
Can you add it there instead?

sbesson: That's the first thing I tried but it didn't work, I can retest. Is is possible that https://github.com/IDR/deployment/blob/master/ansible/group_vars/proxy-hosts.yml#L232 is also required ?

sbesson: Retested the following configuration on `test71`. The redirect is not working . Do you have a suggestion on how to use the redirect map for the `/search` queries?

```diff
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -227,6 +227,9 @@ nginx_proxy_redirect_map:
 - match: ~/pgpc
   dest: /webclient/?show=screen-1151
 
+- match: /search/?query=Name:idr0065
+  dest: /about/download.html
+
 nginx_proxy_redirect_map_locations:
 # TODO: change to 301 when we're happy
 - location: "~ ^/(mito|tara|pgpc)($|/)"
```

sbesson:
Thanks Simon, merging for now as this has been deployed and we will need to remember to revert the redirect in preparation of the publication of `idr0065` /cc @dominikl @francesw

PR #213, opened 2019-11-25
Also add idr0068 redirect to download page

sbesson:


sbesson:
Successfully deployed and tested on `test71`. Proposing to merge and deploy on `prod71` post lunch if no objection.

sbesson:
Deployed to `prod71`7

PR #215, opened 2019-12-05
Add new Protein mapr category

sbesson:
Add new menu for map annotation with openmicroscopy.org/mapr/protein
namespace


sbesson:
Now deployed on `test71` 

Screen Shot 2019-12-05 at 16 40 53

sbesson:
Thanks. Also received confirmation from @francesw she was happy with the Protein mapr category while reviewing `prod72`

PR #216, opened 2019-12-05
Set the top-left webclient logo

manics:
It must've gone missing at some point in the past...


manics:
Deployed on `test71` after merging in https://github.com/IDR/deployment/pull/215 locally
Screen Shot 2019-12-05 at 17 28 45

manics:
Note the IDR logo is not as wide as the default OMERO one, and could be further reduced to just an icon, which would give enough space to add another mapr category to the top-links

PR #217, opened 2019-12-09
Final prod72 adjustmenets

sbesson:
- remove custom redirect to the download page for idr0065
- TBD: update idr.openmicroscopy.org with updated stats


sbesson:
Now released to `prod72`. Merging and tagging the IDR repos.

PR #220, opened 2020-01-07
extcloud01: customisations and bug fixes

manics:
Changes for extcloud01:
- added customisations for things like network interfaces (default unchanged)
- added a firewall (disabled by default but we should consider enabling it on the next prod server too)
- NFS config uses the role and private variables (this is a change relative to `prod73` but should've been done ages ago)
- Fixed SELinux permissions on Nginx cache, requires https://github.com/ome/ansible-role-nginx-proxy/pull/27

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ script:
     --exclude ./roles/kubernetes/
     --exclude ./vendor/
     --exclude ../kubespray/
+    --exclude idr-firewall.yml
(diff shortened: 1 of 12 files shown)


joshmoore:
Since the firewall changes are behind a boolean (`idr_enable_iptables_firewall`) are more or less inert with respect to existing deployments and that merging this is safe. Leaving you to confirm that, then good to merge.

PR #222, opened 2020-02-17
Prod80 final updates

manics:
- Ensures the canonical IDR URL is always used in API responses.
- Bumps iviewer with the projection limit.
- Update about pages (prod80a)
Deployed on `prod80`.

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -134,6 +134,7 @@ omero_web_release: "{{ idr_omero_web_release }}"
 omero_web_config_set:
   # web
   omero.web.api.max_limit: 1000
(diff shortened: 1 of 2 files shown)


will-moore:
If this is on ```prod80```, I still see URLs to "prod80" itself, not to ```idr.openmicroscopy.org```.
E.g. at /api/v0/

manics:
You might be looking at the wrong server, `prod80` is currently down whilst the rsync takes place!

manics:
I've added the iviewer bump to this PR, now live on `prod80`.

sbesson:
The `iviewer` version is as expected. I am unclear which workflow is required  to test the `omero.web.api.absolute_url`. Otherwise LGTM

PR #226, opened 2020-03-11
Add ebi-migration page

manics:
Adds a holding page where idr-analysis can be redirected

Deployed to prod80a, DNS updated


manics:
Done! https://github.com/IDR/repos/releases/tag/0.8.0-1

PR #228, opened 2020-03-31
Update UoD NFS config

manics:
This hasn't been used for ages, we should probably move it to the private config since it's useless for anyone else but that can happen later. For now this updates it to work with the current internal. UoD setup.

### DIFF
--- a/ansible/idr-dundee-nfs.yml
+++ b/ansible/idr-dundee-nfs.yml
@@ -11,6 +11,9 @@
   roles:
   - role: ome.nfs_mount
     nfs_share_mounts:
-    - path: "{{ idr_mountpoint }}"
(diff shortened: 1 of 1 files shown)


joshmoore:
If it's documented somewhere that that's where data is expected, then happy to move the file. Otherwise, it seems like a harmless example of how we have things setup. Otherwise, :+1:

manics:
`/uod/idr` is effectively an internal detail since we haven't published any sort of standardised workflow. If you spin up an empty IDR using these playbooks all you have is the OMERO data directory `/data/OMERO`.

There's a separate file for EBI: https://github.com/IDR/deployment/blob/master/ansible/idr-ebi-nfs.yml so I'd like to get this change in now since we need it and the file already exists, then figure out how to move these two files later.

PR #229, opened 2020-04-07
prod80b OMERO (upgrade)

manics:
OMERO
- server ~5.6.1~ 5.6.0 due to https://github.com/ome/omero-server/issues/93
- py 5.6.2
- web 5.6.3
- includes an option for mounting `BioFormatsCache` as readonly after release to prevent accidental invalidation if there are problems on the public servers.

Requires:
- [x] https://github.com/ome/ansible-role-iptables-raw/pull/3


manics:
This has **not** been applied to `prod80a`, only `prod80b`

ansible/requirements.yml
```diff
@@ -39,8 +39,10 @@
 - src: ome.ice
   version: 4.0.0
 
-- src: ome.iptables_raw
-  version: 0.3.0
+# https://github.com/ome/ansible-role-iptables-raw/pull/3
+- name: ome.iptables_raw
```
sbesson, line 43: Should this be updated to use the new version of the role released to Galaxy?

PR #230, opened 2020-04-13
Consume 0.6.3 release of IDR Bio-Formats

sbesson:
This patch release includes the MetaXpress fix required for idr0081

To be tested in a testing resource first to confirm memo files are not invalidated. Current target release: `prod82`


manics:
Deployed to `prod80b`

PR #236, opened 2020-04-30
Stats for idr0064 idr0083

manics:
Update with https://github.com/IDR/idr.openmicroscopy.org/pull/81


manics:
This is deployed on all current production servers.

PR #237, opened 2020-05-01
Add study-level redirects for figures of idr0083

sbesson:
These URLs pointing to figure panels can be reused in tweets and press releases

Changes have been applied manually in production on `prod80a` and `prod80b` so the main test here is to run the `idr-proxy.yml` in check-mode against either or both servers and test there is no diff (or only whitespaces)

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -236,9 +236,15 @@ nginx_proxy_redirect_map:
 - match: ~/pgpc
   dest: /webclient/?show=screen-1151
 
+- match: /study/idr0083/figure/4i
+  dest: /webclient/img_detail/9822151/?dataset=10201&x=34105&y=84808&zm=25&c=1|318:9927$808080&m=g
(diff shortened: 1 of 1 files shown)


ansible/group_vars/proxy-hosts.yml
```diff
@@ -236,9 +236,15 @@ nginx_proxy_redirect_map:
 - match: ~/pgpc
   dest: /webclient/?show=screen-1151
 
+- match: /study/idr0083/figure/4i
+  dest: /webclient/img_detail/9822151/?dataset=10201&x=34105&y=84808&zm=25&c=1|318:9927$808080&m=g
+  dest: /webclient/?show=screen-1151
```
manics, line 241:
```suggestion
```

sbesson:
Yes I only modified the proxy configuration for the publicly available read-only servers by hand. The changes to `proxy-omeroreadwrite.conf` make sense and I think there should be no harm in deploying them in the upcoming IDR deployment if we wanted to test

manics:
Deployed to `prod80b`

PR #239, opened 2020-05-06
Adjust shell commands for memo regeneration for Bio-Formats.

mtbc:
Small tweaks to suggested shell commands, see diff.

### DIFF
--- a/docs/operating-procedures.md
+++ b/docs/operating-procedures.md
@@ -81,8 +81,8 @@ objects:
     $ /opt/omero/server/OMERO.server/bin/omero login public@localhost -w public
     $ /opt/omero/server/OMERO.server/bin/omero hql --limit -1 --ids-only --style csv 'select MIN(field.image.id) FROM WellSample AS field GROUP BY field.well.plate' > plates.txt
     $ /opt/omero/server/OMERO.server/bin/omero hql --limit -1 --ids-only --style csv 'select d.id from Dataset d' > datasets.txt
-    $ < plates.txt cut -d ',' -f2 | sed -e 's/^/Image:/' > ids.txt
-    $ < datasets.txt cut -d ',' -f2 | sed -e 's/^/Dataset:/' >> ids.txt
+    $ cut -d ',' -f2 plates.txt | sed -e 's/^/Image:/' > ids.txt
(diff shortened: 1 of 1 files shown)


docs/operating-procedures.md
```diff
@@ -80,9 +80,9 @@ objects:
 
     $ /opt/omero/server/OMERO.server/bin/omero login public@localhost -w public
     $ /opt/omero/server/OMERO.server/bin/omero hql --limit -1 --ids-only --style csv 'select MIN(field.image.id) FROM WellSample AS field GROUP BY field.well.plate' > plates.txt
-    $ /opt/omero/server/OMERO.server/bin/omero hql --limit -1 --ids-only --style csv 'select d.id from Dataset d' > datases.txt
-    $ cat plates.txt | cut -d ',' -f2 | sed -e 's/^/Image:/' > ids.txt 
-    $ cat datasets.txt | cut -d ',' -f2 | sed -e 's/^/Dataset:/' >> ids.txt
+    $ /opt/omero/server/OMERO.server/bin/omero hql --limit -1 --ids-only --style csv 'select d.id from Dataset d' > datasets.txt
+    $ < plates.txt cut -d ',' -f2 | sed -e 's/^/Image:/' > ids.txt
```
manics, line 84: Since you're removing the extraneous `cat` you could also remove the `<` by passing the file as a cut argument `cut -d ',' -f2 <file>` 😀

mtbc: True, heh. I kind of like it as it is, the pipeline looks more in order with the input file at the start!

joshmoore: I defer to the more active data importers, but I note remember are using the `$ < ...` idiom anywhere else.

manics: I vote for keeping it simple, bearing in mind we've got several data importers who may not be familiar with advanced shell syntax

mtbc: I'd thought what I'd written was clearer and simpler but will adjust. :laughing:

mtbc:
Travis has actually passed, it's just that GitHub hasn't noticed.

joshmoore:
LGTM (ignoring travis)

PR #243, opened 2020-06-04
Add omero_rois

dominikl:
Adds https://pypi.org/project/omero-rois package, which is handy if one needs to convert segmented images into ROIs.


manics:
Deployed to `prod83`

Issue #245, opened 2020-06-25
Add public credentials to omero-client.json


manics:
- https://github.com/IDR/deployment/blob/master/ansible/templates/omero-client.json.j2
- Need to check this will work as expected with https://github.com/IDR/idr-py/blob/v0.4.0/idr/connections.py#L42

PR #246, opened 2020-06-25
Include public user/pass in omero-client.json

manics:
Closes https://github.com/IDR/deployment/issues/245

Deployed to `test84`

# Testing
```py
from idr import connection
conn = connection('test84.example.org', verbose=2)
for p in conn.listProjects():
    print(p.name)
```


joshmoore:
`http://X/connection/omero-client.json` looks good to me. (We might just need to be ready to scale up ;) )

manics:
I think we should slim down idr-py or extract the connection method if we want everyone to install it, at the moment it pulls in a lot of analysis/pydata/jupyter dependencies

joshmoore:
:+1:

manics:
Or move it upstream 😀
https://github.com/ome/design/issues/106

manics:
Deployed to `prod84`

PR #248, opened 2020-06-30
Bump omero-py version to 5.7.1

sbesson:
See https://github.com/ome/omero-py/blob/v5.7.1/CHANGELOG.md for the corresponding changes

Target: `prod85`


sbesson:
Deployed on `test84`

manics:
Discussed in slack, merging.

PR #251, opened 2020-07-18
Bump omero-cli-render to version 0.6.0

sbesson:
See https://github.com/ome/omero-cli-render/blob/v0.6.0/CHANGES.md for the list of changes


sbesson:
Tested while rendering the plates of `idr0078`. Proposing to get this merged for `prod86`

PR #252, opened 2020-07-21
Update static website for upcoming prod85 release

sbesson:
To be deployed on `prod85`


sbesson:
Deployed on `prod85`. Ready for public release

PR #256, opened 2020-07-22
FTP and file transfer command line utils

manics:
Installs cli-utils and rclone.


manics:
Deployed to idrftp

manics:
Also added the aspera transfer script.

PR #260, opened 2020-08-19
Add pandas to notebook

pwalczysko:
During testing of https://github.com/IDR/idr-utils/pull/11 it turned out that no pandas are installed on idr-next.

This PR is adding the line to the playbook.
The command run manually as 

``/path-to/pip install pandas``

showed the version to be 1.1.0

@manics


dominikl:
Thanks Petr, forgot about that! Hope it won't be an issue to add it to the default IDR server venv.

manics (approved):

PR #263, opened 2020-08-25
Bump omero-iviewer to version 0.10.1

sbesson:
This is primarily driven by the  timestamp bug fix in iviewer. As per the dependency, it also pulls omero-web 5.7.0

See 
- https://github.com/ome/omero-iviewer/blob/v0.10.1/CHANGELOG.md
- https://github.com/ome/omero-web/blob/v5.7.0/CHANGELOG.md

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -212,7 +212,7 @@ omero_web_config_set:
 
 omero_web_apps_packages:
 - omero-mapr==0.4.0
-- omero-iviewer==0.9.1
+- omero-iviewer==0.10.1
 - omero-gallery==3.3.3
 omero_web_apps_names:
 - omero_mapr
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -13,8 +13,8 @@ omero_server_checkupgrade_comparator: '!='
 
 idr_omero_web_release: 5.7.0
 # omero-web depends on omero-py but may not pin the latest release
-# omero_web_python_addons:
-#   - omero-py==5.7.1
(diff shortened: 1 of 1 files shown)


sbesson:
Deployed on `pilot-idr0092`

ansible/group_vars/omero-hosts.yml
```diff
@@ -11,10 +11,10 @@
 idr_omero_server_release: 5.6.0
 omero_server_checkupgrade_comparator: '!='
 
-idr_omero_web_release: 5.6.3
+idr_omero_web_release: 5.7.0
 # omero-web depends on omero-py but may not pin the latest release
-omero_web_python_addons:
-  - omero-py==5.7.1
+# omero_web_python_addons:
+#   - omero-py==5.7.1
```
manics, line 17: Is there a reason for commenting this out?

sbesson: No specific reason, this might come from debugging (as bumping `iviewer` alone was not sufficient since upstream `omero-web` versions are required.
I assume we want to keep tracking the individual `omero-web/omero-py` versions in the deployment rather than letting `omero-web` pull the latest compatible dependency on deployment?

manics: Yes, that's my preference until we have a [better way of describing dependencies](https://github.com/ome/design/issues/103).

sbesson: Restored in  230267c

manics (approved):
LGTM. Are you deploying it for the next release?

sbesson:
Yes, I'll deploy on `prod87` once merged

sbesson:
Deployed on `prod87`

PR #266, opened 2020-09-03
Decommissioning improvements

sbesson:
Small improvements:

- following a RFE from @joshmoore, ebabdd3  should ensure monitoring and slack alerting agents are shut down before stopping the other services
- 28f6ec6 primarily aims at speeding up the decommissioning process. At the moment, the DB dump/log archives are first downloaded locally then copied over to our UoD location. Skipping the local copy step, I ended up simply rsyncing the logs/dumps into their final location on `idr0-slot3` as:

   ```
   rsync -av idr-next:/tmp/prod86-proxy-* . --progress
   rsync -av -e "ssh idr-next ssh" database:/tmp/prod86-database-* . -vv --progress
   rsync -av -e "ssh idr-next ssh" management:/tmp/prod86-management-* . --progress
   ```

If the above sounds like the canonical to go, happy to more simply remove the `fetch` commands.

With the transfer step reduced to a single copy step, I think most of the bottleneck of the decommissioning process comes down to the `pgdump` itself (currently dumping ~5GB takes typically 1h). If the dump format is not critical, the obvious next thing to try would be to switch to parallel export with `-Fd` in https://github.com/IDR/deployment/blob/IDR-0.8.7/ansible/decommission/archive-db.yml#L28

### DIFF
--- a/ansible/decommission/archive-instance-services.yml
+++ b/ansible/decommission/archive-instance-services.yml
@@ -12,15 +12,15 @@
       state: stopped
       enabled: False
     with_items:
+    - prometheus-node-exporter
+    - prometheus-omero-py
+    - omero-logmonitor
     - omero-server
     - omero-web
-    - omero-logmonitor
     - redis
     - httpd
     - nginx
     - postgresql-9.6
     - postgresql-11
-    - prometheus-node-exporter
-    - prometheus-omero-py
     - docker
     ignore_errors: yes
--- a/ansible/decommission/archive-instance-services.yml
+++ b/ansible/decommission/archive-instance-services.yml
@@ -12,8 +12,6 @@
       state: stopped
       enabled: False
     with_items:
-    - prometheus-node-exporter
-    - prometheus-omero-py
     - omero-logmonitor
     - omero-server
     - omero-web
@@ -22,5 +20,7 @@
     - nginx
     - postgresql-9.6
     - postgresql-11
+    - prometheus-node-exporter
+    - prometheus-omero-py
     - docker
     ignore_errors: yes
--- a/ansible/decommission/archive-db.yml
+++ b/ansible/decommission/archive-db.yml
@@ -46,6 +46,7 @@
       src: /tmp/{{ decommission_archive_prefix }}-database-{{ item.name }}.pg_dump
     with_items:
     - "{{ postgresql_databases }}"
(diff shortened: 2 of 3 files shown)


ansible/decommission/archive-instance-services.yml
```diff
@@ -12,15 +12,15 @@
       state: stopped
       enabled: False
     with_items:
+    - prometheus-node-exporter
+    - prometheus-omero-py
+    - omero-logmonitor
```
manics, line 17: Did this do what you wanted it to? I'd have thought that you'd get alerts due to those agents going down whilst the central monitoring server remained up.

sbesson: I have not and it is possible that the order is incorrect. The decommissioning of `prod86` led to a large number of alerts of type

```
*: prod86-omeroreadonly-3 /opt/omero/server/OMERO.server/var/log/master.err:
-! 9/3/20 12:01:09:480 warning: Blitz-0-Ice.ThreadPool.Server-13: dispatch exception:
   identity: session-06b53b7b-eff9-4c32-93e2-a929c4d91f6e/1a6fc957-ed03-413f-9282-5ee8ca4dca41
   facet: 
   operation: destroy
```

Maybe just stop `omero-logmonitor` before stopping the OMERO services?

sbesson:
> In line with our general aim of reducing manual steps it'd be good if there was a script to wrap the rsync commands.

Agreed and happy to start working on this if we are fine with this approach. Would https://docs.ansible.com/ansible/latest/modules/synchronize_module.html be the starting point for a remote-to-remote rsync?

ansible/decommission/archive-instance-services.yml
manics: Yes, I think just stopping omero-logmonitor will reduce some of the log alerts, though probably not all since we also have fluentd.

manics:
For the script I was thinking more of a shell script on idr0-slot3 rather than ansible. I think trying to synchronise between two remote servers, both behind an ssh gateway, is going to be painful in Ansible, though it would be pretty neat if it worked.

ansible/decommission/archive-instance-services.yml
sbesson: :+1: pushed 077d0c8

manics (approved):

PR #267, opened 2020-09-03
Deploy 2020-09-03 version of the website including updated stats

sbesson:
This includes the stats for `prod86` and `prod87` and has been deployed in production - see idr.openmicroscopy.org/about/studies.html


sbesson:
Merging and tagging as `IDR-0.8.7-1`

PR #268, opened 2020-09-08
Add redirects for idr0078, idr0089, idr0090 and idr0094

sbesson:
These studies are scheduled for publication in upcoming releases of IDR. In the meantime, the searches should redirect to the download page as the raw data is publicly available.

To be deployed on `idr-next` for testing, then on `idr`

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -295,7 +295,14 @@ _nginx_proxy_sites:
     # Study redirects
     - "if ($request_uri ~ /search/\\?query=Name:idr0068) {
        return 302 /about/download.html;}"
-
+    - "if ($request_uri ~ /search/\\?query=Name:idr0078) {
+       return 302 /about/download.html;}"
+    - "if ($request_uri ~ /search/\\?query=Name:idr0089) {
+       return 302 /about/download.html;}"
+    - "if ($request_uri ~ /search/\\?query=Name:idr0090) {
+       return 302 /about/download.html;}"
+    - "if ($request_uri ~ /search/\\?query=Name:idr0094) {
+       return 302 /about/download.html;}"
 
   # This is a duplicate of the main OMERO.web proxy configuration, but with
   # cache-busting:
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -293,15 +293,15 @@ _nginx_proxy_sites:
     nginx_proxy_additional_directives:
     - "add_header Access-Control-Allow-Origin $allow_origin"
     # Study redirects
-    - "if ($request_uri ~ /search/\\?query=Name:idr0068) {
+    - "if ($request_uri ~ /search/\\?query=Name:(.*)68) {
        return 302 /about/download.html;}"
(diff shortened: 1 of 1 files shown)


sbesson:
Now deployed on `idr-next` for testing.

sbesson:
Pushed an extra commit and deployed it on `idr-next` to support variants of the name search e.g. https://idr-next.openmicroscopy.org/search/?query=Name:idr0068, https://idr-next.openmicroscopy.org/search/?query=Name:0068 or https://idr-next.openmicroscopy.org/search/?query=Name:68  should all redirect to the downloads page now

francesw (approved):
Looks good, all redirects working.

francesw (approved):

ansible/group_vars/proxy-hosts.yml
```diff
@@ -293,9 +293,16 @@ _nginx_proxy_sites:
     nginx_proxy_additional_directives:
     - "add_header Access-Control-Allow-Origin $allow_origin"
     # Study redirects
-    - "if ($request_uri ~ /search/\\?query=Name:idr0068) {
+    - "if ($request_uri ~ /search/\\?query=Name:(.*)68) {
```
manics, line 296: At some point we'll end up with `idr0168` and this will break, can the regex check for something like `0*`?

sbesson: The trouble is that URLs of type https://idr.openmicroscopy.org/search/?query=Name:68 are now being used in the wild and we need these to redirect until the data is published.

I have raised your exact issue with the increasing accession numbers. For now I'll deploy but happy to have this discussed on weekly call.

sbesson:
Depoyed to `idr`. I'll merge and tag. We can discuss the URL shortening separately as I agree the variant might cause us trouble down the line (although I would hope we could have published all `idr00xx` studies by then).

ansible/group_vars/proxy-hosts.yml
sbesson: e14aef4 should now constraint the regex to only match the two known variants of the search URLs i.e.
 https://idr-next.openmicroscopy.org/search/?query=Name:idr0068 and https://idr-next.openmicroscopy.org/search/?query=Name:68. As suggested, https://idr-next.openmicroscopy.org/search/?query=Name:168 should now return no found study as expected

francesw: Instead of using the wildcard if statement, which could cause a problem with 168 etc, can we take search term and remove any non-numeric characters then convert to an integer which would make 68 unique?

sbesson: @francesw can you retest the latest version deployed on `idr-next` which should only redirect `Name:68` and `Name:idr0068` ?

francesw: Tested, works as expected.

manics: @francesw I think a good longer term fix is to handle this as a new OMERO.web app where we've got the full manipulative of Python, instead of doing this in Nginx.

sbesson:
Agreed with a revamp of the searching/redirecting  logic via the app eventually. Merging, tagging and deploying to `prod87` immediately

PR #269, opened 2020-09-14
Add idr-s3.openmicroscopy.org readonly public proxy for /idr

manics:
Changes the `idr-ftp` `idr-upload` read-write authenticated proxy to only apply to virtualhost `idr-ftp.openmicroscopy.org`.
All other hostnames will be redirected to https://idr.openmicroscopy.org/

### DIFF
--- a/ansible/idr-s3gateway.yml
+++ b/ansible/idr-s3gateway.yml
@@ -7,7 +7,7 @@
   roles:
 
     - role: ome.minio_s3_gateway
-      minio_s3_gateway_remote_endpoint: "https://s3.embassy.ebi.ac.uk"
+      minio_s3_gateway_remote_endpoint: "{{ idr_s3_gateway_remote_endpoint }}"
       minio_s3_gateway_bucket: "{{ idr_minio_s3_gateway_bucket | default('example') }}"
       minio_s3_gateway_access_key: "{{ idr_secret_minio_s3_gateway_access_key | default('minio') }}"
       minio_s3_gateway_secret_key: "{{ idr_secret_minio_s3_gateway_secret_key | default('minio123') }}"
@@ -48,7 +48,7 @@
           nginx_proxy_backends:
             - name: s3-embassy-idr
               location: "~ ^/idr(/.*)?$"
-              server: "https://s3.embassy.ebi.ac.uk"
+              server: "{{ idr_s3_gateway_remote_endpoint }}"
           # Disable buffering
           # https://serverfault.com/a/818090
           nginx_proxy_additional_directives:
@@ -72,3 +72,6 @@
       service:
         name: nginx
         state: restarted
+
+  vars:
+    idr_s3_gateway_remote_endpoint: "https://s3.embassy.ebi.ac.uk"
--- a/ansible/idr-s3gateway.yml
+++ b/ansible/idr-s3gateway.yml
@@ -28,17 +28,41 @@
       nginx_proxy_http2: true
       nginx_proxy_force_ssl: true
       # nginx_proxy_hsts_age: 31536000
-      nginx_proxy_backends:
-        - name: s3gateway
-          location: "~ ^/{{ idr_minio_s3_gateway_bucket | default('example') }}(/.*)?$"
-          server: "http://localhost:9000"
-      # Disable buffering and allow unlimited file sizes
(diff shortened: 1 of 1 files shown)


ansible/idr-s3gateway.yml
```diff
@@ -28,17 +28,41 @@
...
+        # idr-ftp.openmicroscopy.org: S3 read-write /idr-upload only
+        - nginx_proxy_server_name: idr-ftp.openmicroscopy.org
+          nginx_proxy_backends:
+            - name: s3gateway
+              location: "~ ^/{{ idr_minio_s3_gateway_bucket | default('example') }}(/.*)?$"
+              server: "http://localhost:9000"
+          # Disable buffering and allow unlimited file sizes
+          # https://serverfault.com/a/818090
+          nginx_proxy_additional_directives:
+            - client_max_body_size 0
```
joshmoore, line 41: This is only needed for the PUTs not the GETs below?

ansible/idr-s3gateway.yml
```diff
@@ -28,17 +28,41 @@
...
-          server: "http://localhost:9000"
-      # Disable buffering and allow unlimited file sizes
-      # https://serverfault.com/a/818090
-      nginx_proxy_additional_directives:
-        - client_max_body_size 0
-        - proxy_http_version 1.1
-        - proxy_request_buffering off
-        - proxy_buffering off
+      nginx_proxy_sites:
+        # idr-ftp.openmicroscopy.org: S3 read-write /idr-upload only
```
joshmoore, line 32: Not specific to this PR but `idr-ftp...` for s3 seems inflammatory ;)

ansible/idr-s3gateway.yml
```diff
@@ -28,17 +28,41 @@
...
+        - nginx_proxy_server_name: idr-ftp.openmicroscopy.org
+          nginx_proxy_backends:
+            - name: s3gateway
+              location: "~ ^/{{ idr_minio_s3_gateway_bucket | default('example') }}(/.*)?$"
+              server: "http://localhost:9000"
```
joshmoore, line 37: Also not overly specific to this PR, but having the bucket as a variable but not the server seems somewhat odd.

joshmoore (approved):
Looks generally good. I think we're stretched in terms of how re-usable this file is in other situations (e.g. all the instances of `openmicroscopy.org`) but I'm all for getting this in and then we can figure out how to turn it into infrastructure.

ansible/idr-s3gateway.yml
manics: Correct, the body of a client GET request should be very small.

ansible/idr-s3gateway.yml
manics: No one came up with a better name 😄 .

Though now that we have idr-s3 I think we could put it under there, providing we never need to split the two buckets onto different servers.

ansible/idr-s3gateway.yml
manics: It's probably a bit confusing, because its an Nginx proxy to the S3 minio proxy to the real s3 embassy server

joshmoore: I don't think that's what triggered me, but more it's unclear what gets encoded directly in the file and what is a variable.

joshmoore:
:+1: I'd say let's go for it and I'll start pointing people at it to see if it falls over ;)

ansible/idr-s3gateway.yml
sbesson: Although we extensively discussed the `idr-s3` proxy server for the public Zarr data, I do not remember discussing the proxy server name for the `idr-upload` bucket. I had been probably wrongly `idr-s3` could be reused as well.

I also feel `idr-ftp.openmicroscopy.org/idr-upload` is confusing for submitters esp. if we keep maintaining a FTP service in parallel. As the scope of this PR is the public read-only `idr` bucket, I would not raise as a blocker but let's not sidestep this discussion before we start pointing submissions at the upload bucket.

joshmoore: https://github.com/IDR/deployment/issues/270

joshmoore (approved):
Still looks good.

PR #272, opened 2020-09-16
Remove redirect for idr0078

sbesson:
In preparation of the final checks, removes the top-level redirect to the downloads page


sbesson:
Deployed on `prod88`

francesw:
I still see the redirect on idr-next (prod88).

francesw:
I still see the redirect on idr-next (prod88).

sbesson:
Sorry, I must have reverted the change as part of the work to fix redirects on #273. Redeployed to `prod88` now

PR #274, opened 2020-09-16
OMERO.web 5.8.0 OMERO.py 5.8.0

manics:
Not deployed anywhere, we can wait for the next test server


sbesson (approved):
Any reason not to deploy on `idr-next` / `prod88` immediately? I need to run the playbooks to update the top-level redirects today

manics:
This will cause an OMERO.server restart, so needs to be co-ordinated with data importers.

sbesson:
Deployed on `prod88`

PR #276, opened 2020-09-28
Drop addition of Access-Control header

joshmoore:
Having the header added twice broke clients. Removing
for the moment, but a conditional addition would be
safest.

see: https://forum.image.sc/t/mirroring-idr-zarr-datasets/41136/16?u=joshmoore

### DIFF
--- a/ansible/idr-s3gateway.yml
+++ b/ansible/idr-s3gateway.yml
@@ -55,9 +55,6 @@
             - proxy_http_version 1.1
             - proxy_request_buffering off
             - proxy_buffering off
-            # CORS: basically allow any cross-site since this is public
-            # read-only. Always set a header including on 404 responses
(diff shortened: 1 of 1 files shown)


joshmoore:
@manics : I commented out the addition on idr-s3. Happy to run this to remove it unless you have an idea on how to approach the conditionalism.

manics:
Why don't we redirect to s3.embassy instead?

joshmoore:
Shall I just close this then?

manics:
Might as well leave it, I'll mark it closed with the PR that adds the redirect.

joshmoore:
Ta.

manics:
Merging since this needs to be redeployed to update the SSL cert. In hindsight should've merged this earlier since it was deployed to production even if it's not the best solution.

PR #277, opened 2020-09-28
Update static website

sbesson:
See https://github.com/IDR/idr.openmicroscopy.org/compare/2020.09.03...2020.09.28 for the list of changes.

Primarily stats additions for `prod88`


sbesson:
Now deployed on `prod88`

PR #278, opened 2020-09-29
Set omeroreadwrite web timeout to 900

manics:
This is needed to re-cache mapr Homo sapiens queries


manics:
Deployed to `prod88`

PR #279, opened 2020-09-29
Remove unused omeroreadwrite:8009 proxy

manics:
This configuration is unused, everything can be done using port 9000


sbesson (approved):
Deployed on `prod89`. Port 9000 remains functional and I could not find any instructions related ot port 8009. Merging and we can review during the caching step if something is not working.

PR #280, opened 2020-09-30
Redirect idr-s3.openmicroscopy.org → s3.embassy.ebi.ac.uk

manics:
Upstream CORS issue is fixed

Closes https://github.com/IDR/deployment/pull/276

Not yet deployed


joshmoore:
This seems fine, but I don't know if all implementations will behave correctly. I imagine we should deploy somewhere/somehow and then I'm happy to test a few. (awscli, fsspec, mc)

Issue #264, opened 2020-08-25
Add monitoring for pilot IDRs and ancillary servers


manics:
Add prometheus monitoring for pilot-proxy, all pilot IDRs (omeroreadwrite) and idrftp. This will require:
- [ ] Setup a management server that's connected to both the idrftp and pilot-idr networks. This will probably be part of the idr-proxy deployment since the proxy is already connected to all IDRs and is updated as part of the standard pilot deployment workflow. Therefore idr-ftp will require a change to it's networking and configuration to fit in but this shouldn't be too bad.
- [ ] Customising the current omeroreadwrite monitoring configuration to work with the pilot IDRs. This may not be too difficult if the management VM is on the pilot-idr network.
- [x] Short term fix: add disk space alerting to idrftp since this is the priority https://github.com/IDR/deployment/pull/265
- [ ] Add monitoring to idrftp
- [ ] Add alerting to the pilot management VM. Since this is a newish deployment we can use this as a way to experiment with the prometheus alerting rules.

Also
- [ ] Update main IDR deployments with new prometheus and prometheus-node roles. The latter may require Grafana dashboards to be updated since some metric names have changed. https://github.com/IDR/deployment/pull/281
- [x] Update postresql prometheus exporter (currently broken) https://github.com/IDR/deployment/issues/244

joshmoore:
Along with #265, @manics, would it be possible to have a definitive list of monitoring ports and optimally URLs that one can use like http://localhost:9090/graph?g0.range_input=1h&g0.expr=(node_filesystem_avail_bytes%7Bmountpoint%3D%22%2Fdata%22%7D%20*%20100)%20%2F%20node_filesystem_size_bytes%7Bmountpoint%3D%22%2Fdata%22%7D&g0.tab=0 ?

manics:
Other prometheus roles to be upgraded:
- https://github.com/ome/ansible-role-omero-web-django-prometheus/pull/4
- https://github.com/ome/ansible-role-prometheus-jmx/pull/4

PR #281, opened 2020-09-30
Main IDR monitoring upgrade

manics:
Upgrades prometheus and grafana.

Partly related to https://github.com/IDR/deployment/issues/264

The main breaking changes are changes to metric names, which is why some of the Grafana dashboards are modified.

### DIFF
--- a/ansible/grafana-dashboards/idr-per-server.json
+++ b/ansible/grafana-dashboards/idr-per-server.json
@@ -78,7 +78,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "(1 - node_filesystem_free{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"} / node_filesystem_size{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"}) * 100",
+              "expr": "(1 - node_filesystem_free_bytes{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"} / node_filesystem_size_bytes{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"}) * 100",
               "format": "time_series",
               "intervalFactor": 2,
               "legendFormat": "{{mountpoint}} ({{device}})",
--- a/ansible/grafana-dashboards/idr-vertical.json
+++ b/ansible/grafana-dashboards/idr-vertical.json
@@ -80,7 +80,7 @@
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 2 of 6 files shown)


ansible/grafana-dashboards/idr-vertical.json
```diff
@@ -428,12 +428,10 @@
       {
         "allValue": null,
         "current": {
-          "tags": [],
-          "text": "omero + database + docker",
+          "text": "omero + database",
           "value": [
             "omero",
-            "database",
-            "docker"
```
joshmoore, line 436: Where has docker gone?

ansible/grafana-dashboards/idr-per-server.json
```diff
@@ -78,7 +78,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "(1 - node_filesystem_free{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"} / node_filesystem_size{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"}) * 100",
+              "expr": "(1 - node_filesystem_free_bytes{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"} / node_filesystem_size_bytes{fstype!~\"(nfs|nfs4|overlay|rootfs|rpc_pipefs|tmpfs)\", instance=\"$hostname\"}) * 100",
```
joshmoore, line 81: What's the impact of the change? Does that mean that previous values are no longer chartable?

joshmoore (commented):
As general, my worry when I hear "breaking" is that we won't be able to ask historical questions. Can you explain why these happened?

Otherwise, :+1: for keeping monitoring happy & healthy.

ansible/grafana-dashboards/idr-vertical.json
manics: It's been gone for years, ever since we separated the deployment of the VAE from IDR.

ansible/grafana-dashboards/idr-per-server.json
manics: You'd have to create two charts, one for each metric.

manics:
The breaking changes are due to upstream standardisation of the metric names:
- https://github.com/prometheus/node_exporter/blob/v0.16.0/docs/V0_16_UPGRADE_GUIDE.md
- https://github.com/prometheus/node_exporter/releases/tag/v0.16.0

ansible/grafana-dashboards/idr-vertical.json
joshmoore: But isn't it coming back with micro services? (and idr-ftp?)

manics: These are used for selecting a subset of hostnames, it's not actually anything to do with what's running on the system. We don't have nodes named `.*docker.*`, the current microservice works run the docker micoservices on the `omero*` servers since some of them require filesystem access anyway.

manics:
Any more comments, or shall we merge this?

PR #282, opened 2020-10-02
Bump omero-py to 5.8.2 and omero-web to 5.8.1

sbesson:
See https://github.com/ome/omero-py/blob/v5.8.2/CHANGELOG.md and
https://github.com/ome/omero-web/blob/v5.8.1/CHANGELOG.md for the list of changes

Fixes https://github.com/IDR/idr-utils/issues/27

With this change `omero` commands piped to an output file requiring an interactive user input should fail rather than hang.


sbesson:
Deployed on `pilot-idr0094`

```
[sbesson@pilot-idr0094-omeroreadwrite ~]$ /opt/omero/server/venv3/bin/omero login > out
stdout is not a terminal: cannot request server
```

manics (approved):

PR #285, opened 2020-10-05
omero-client.json: use wss (websockets)

manics:
Once this is deployed
```py
from idr import connection()
c = connection('idr.openmicroscopy.org')
```
should automatically use websockets.

Requires https://github.com/IDR/deployment/pull/287


sbesson (approved):
As discussed last week, merging for deployment into `prod91` /cc @jburel

PR #286, opened 2020-10-05
Deploy 2020.10.05 version of idr.openmicroscopy.org/about

sbesson:
Deployed on `prod89`


manics (approved):
image

Maybe we should re-order those in a future release? Without the release name in the table they look out of place.

sbesson:
> Maybe we should re-order those in a future release? Without the release name in the table they look out of place.

I was assuming the Jekyll liquid was already doing so but apparently not. Ordering by the first column makes complete sense to me.

manics:
I assume we can do it in jekyll, though we'd eventually run into https://github.com/ome/design/issues/108

PR #287, opened 2020-10-07
Block non-websocket connections to OMERO.server

manics:
See https://github.com/ome/ansible-role-nginx-proxy/pull/29


jburel (approved):
changes make sense

PR #289, opened 2020-10-07
Also mount EBI NFS mounts on FTP server

sbesson:
An increasing number of IDR submissions are undergoing conversion via `bioformats2raw` esp. to construct pyramidal level. This conversion usually involves:

- access to the original data, minimally in read-only
- some conversion tools/pipeline - Docker-based containers have been predominantly used lately
- writing the converted data to disk i.e. sufficient disk space depending on the data
- copying the local data to EBI NFS - Aspera within EBI is currently the fastest way to achieve this

In order to progress with a submission blocked at the conversion level, this PR proposes to mount EBI NFS as read-only on the FTP server as it meets most of the criteria defined above.

### DIFF
--- a/ansible/idr-ftp.yml
+++ b/ansible/idr-ftp.yml
@@ -23,7 +23,7 @@
   - role: ome.nfs_mount
     nfs_version: 3
     nfs_mount_opts: timeo=14,intr,nolock,ro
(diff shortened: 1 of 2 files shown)


sbesson:
Closing as I wasn't able to update the test FTP server and we will be investigating other options.

PR #291, opened 2020-10-21
Install dstat on idr-ftp

manics:
This is useful for monitoring performance

We could install this on all VMs, for now it's just idr-ftp since we're investigating S3 performance issues.

### DIFF
--- a/ansible/idr-ftp-monitoring.yml
+++ b/ansible/idr-ftp-monitoring.yml
@@ -21,3 +21,10 @@
       iface: "{{ idr_net_iface | default('eth0') }}"
 
     prometheus_additional_rules_template: idrftp-alert-rules-yml.j2
+
+  tasks:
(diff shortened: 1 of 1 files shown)


manics:
This isn't actually needed since we can get everything from prometheus `http://localhost:9090/graph?g0.range_input=1h&g0.expr=rate(node_network_transmit_bytes_total%5B2m%5D)&g0.tab=0&g1.range_input=1h&g1.expr=rate(node_network_receive_bytes_total%5B2m%5D)&g1.tab=0`

But it's installed now...

sbesson:
> But it's installed now...

So are you in favor of merging this to capture the current state or closing this PR as this extra utility is not required?

manics:
Merge it on the basis that it's better to record the actual state of our systems rather than what we'd like it to be.

sbesson (approved):

PR #292, opened 2020-10-29
Redirect /curation-workflow

manics:
Final location is still to be decided


manics:
Deployed to `test90`

joshmoore (approved):
Looks ok to me (even though the latest change hasn't been deployed: still redirects to github)

One thought: this does change the URL which means that users may book mark about/curation-workflow rather than /curation-workflow meaning we may need to redirect _both_.

manics:
Deployed to all three deployments.

sbesson (approved):

sbesson:
Merging despite this PR being still marked as draft. Can you push a tag to mark the fact this has been deployed to `prod89`?

manics:
Sorry, didn't realise it still had a draft label. I'll push a tag `IDR-0.8.9-1`

PR #294, opened 2020-11-16
Remove mineotaur

manics:
The trigger for this is that the OMERO servers will soon also be `dockermanager-hosts` to help with running microservices. The NFS configuration in for `dockermanager-hosts` overwrites the OMERO NFS configuration, and since Mineotaur is no longer working we might as well remove it.


sbesson (approved):
As discussed this will be deployed on `prod91`

PR #297, opened 2020-11-17
Add all microservices

manics:
- omero-ms-thumbnail 0.5.2
- omero-ms-imageregion 0.5.2 (https://github.com/manics/docker-example-omero-microservices/blob/1549d55923637bb34e3f047f18b17fe5233caaa8/imageregion/Dockerfile, manually built and pushed to Docker Hub)
- omero-ms-zarr 0.1.6

The front-end Nginx configuration is in a separate commits in case we want to split this PR and deploy the microservices internally without exposing them.

omero-ms-imageregion and omero-ms-zarr are currently broken for anything that requires the IDR fork of BioFormats.

BioFormatsCache is read-only inside the docker images, so if you open a large image it will probably time out. Update: presumably the version of BioFormats in the micro-services will use a different memo version anyway, so even if it was writeable it'd delete the old cache?


ansible/group_vars/proxy-hosts.yml
```diff
@@ -292,9 +349,11 @@ _nginx_proxy_sites:
   # This enables the default site (configured using the global
   # nginx_proxy_* variables):
   - nginx_proxy_is_default: True
-    # Only add this CORS header to the default public site
+
     nginx_proxy_additional_directives:
-    - "add_header Access-Control-Allow-Origin $allow_origin"
+    # Only add this CORS header to the default public site
+    # Always set a header including on 404 responses
+    - "add_header Access-Control-Allow-Origin $allow_origin always"
```
manics, line 356: Needed for vizzarr

ansible/idr-firewall.yml
```diff
@@ -12,7 +12,7 @@
       become: yes
       iptables_raw_25:
         name: default_accept
-        keep_unmanaged: no
+        keep_unmanaged: yes
```
manics, line 15: Needed because Docker doesn't play well with fixed iptables rules.

ansible/templates/omero-ms-imageregion-entrypoint.sh.j2
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -eu
+
+grep imageregion /etc/group || addgroup -g {{ idrnfs_groupid | default(1000) }} imageregion
```
manics, line 4: This is needed due to how NFS is setup

ansible/templates/omero-ms-zarr-entrypoint.sh.j2
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+set -eu
+
+grep omerozarr /etc/group || addgroup -g {{ idrnfs_groupid | default(1000) }} omerozarr
```
manics, line 4: This is needed due to how NFS is setup

joshmoore: Might be worth capturing this and other tidbits in the PR itself.

PR #300, opened 2020-12-08
Replace travis molecule with GitHub workflow

manics:
Replaces Travis with a GitHub Workflow.

The version of molecule used in https://github.com/ome/ome-ansible-molecule/tree/v0.4.4 depends on an old version of testinfra which has a bug, so we need to manually upgrade it in the workflow to get the tests to pass.


sbesson (approved):
https://github.com/manics/deployment/actions/runs/463945928 is green so let's get this merged so that new PRs can be tested and reviewed

Re https://github.com/IDR/deployment/pull/300/commits/1f7a1de7578f0417465f743736525ac79fa4a233, is there a follow-up action. Should the working version of `test-infra` be pinned in a release of `ome-ansible-molecule` or do you think upgrading Molecule would be doable?

manics:
I don't know whether pip will reliably handle conflicting dependencies in the same installation step. I haven't look at upgrading molecule, though it'll be needed to support newer Ansible versions.

PR #302, opened 2020-12-14
OMERO.py 5.9.0rc1: disable unused OMERO components

manics:
Disable all components except Blitz and Tables

Fixes the postgres grafana dashboard (was exported with the wrong options in https://github.com/IDR/deployment/pull/299)

### DIFF
--- a/ansible/grafana-dashboards/idr-postgresql.json
+++ b/ansible/grafana-dashboards/idr-postgresql.json
@@ -1,4 +1,34 @@
 {
+  "__inputs": [
+    {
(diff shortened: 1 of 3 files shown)


sbesson (approved):
Thanks. The removal of the pilot configuration makes sense. I assume this only affected the default view (tested on `prod91` vs `prod92`).

All other changes were discussed independently.

I had to restart Travis CI a few times. I suspect we will want to complete #300 soon in order to merge subsequent PRs.

PR #306, opened 2021-01-13
Update monitoring docs

manics:
Adds a bit more information to the IDR monitoring docs.

### DIFF
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -1,6 +1,6 @@
 # IDR Monitoring
 
-Each IDR deployment has it's own monitoring system.
+Each IDR deployment has its own monitoring system.
 
 ## Metrics
 
--- a/docs/monitoring.md
+++ b/docs/monitoring.md
@@ -1,5 +1,7 @@
 # IDR Monitoring
 
+Each IDR deployment has it's own monitoring system.
+
 ## Metrics
 
 Day to day performance of the IDR can be visualised using Grafana, which is accessible using an anonymous account on idr-management:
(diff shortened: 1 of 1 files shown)


docs/monitoring.md
```diff
@@ -1,5 +1,7 @@
 # IDR Monitoring
 
+Each IDR deployment has it's own monitoring system.
```
joshmoore, line 3:
```suggestion
Each IDR deployment has its own monitoring system.
```

docs/monitoring.md
```diff
@@ -32,3 +34,43 @@ Limitations:
...
+
+## IDR FTP Monitoring
+
+The IDR FTP server has limited Prometheus monitoring running on the same server with alerts for low available disk space.
+
+
+## IDR Pilot Monitoring
+
+All IDR pilot test instances are currently monitored by a single monitoring server.
+This is accessed using the same commands as above, using `idr-pilot` as the proxy.
```
joshmoore, line 47: Might be better to be explicit since there's the "centralized logged" between this and the first addition. (Or move the logging to the bottom?)

docs/monitoring.md
```diff
@@ -32,3 +34,43 @@ Limitations:
...
+
+
+## Deployment
```
joshmoore, line 50: Looking at the outline (i.e. `##` markdown level) you're trying to say here, "this is for the people setting it up" and the bits above are more for the "users"?

joshmoore (commented):
Few minor thoughts.

docs/monitoring.md
manics: Yes, I've re-arranged the doc slightly, is that clearer?

PR #307, opened 2021-01-13
Add idr-s3 docs

manics:
Add a short doc about the idr-ftp S3 Minio upload server

### DIFF
--- a/docs/idr-s3.md
+++ b/docs/idr-s3.md
@@ -1,0 +1,21 @@
+# IDR S3 upload server
+
+The IDR S3 submission server is co-located with the IDR FTP server.
+The IDR S3 server is actually a proxy through to the backend S3 filestore provided by EBI, and uses Minio's support for proxying S3 whilst overlaying its own authentication.
+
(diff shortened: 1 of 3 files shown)


joshmoore (approved):
:+1:

docs/idr-s3.md
```diff
@@ -0,0 +1,21 @@
+# IDR S3 upload server
+
+The IDR S3 submission server is co-located with the IDR FTP server.
```
sbesson, line 3: This made me wonder: we have no similar documentation describing the FTP server deployment?

sbesson (approved):
Thanks, overall makes sense. Eventually, I can also see this document cross-linked from the submission workflow as we will need to create Minio users per submission so it is good to have a starting page.

One general question re the layout of these docs. It looks like originally the documents were close the deployment procedure (`provisioning`, `deployment`). More recently, we introduced documents per service: `monitoring`, `s3`. Would a top-level index e.g. a table of contents for the various topics covered by these docs help the discoverability?

manics:
I've renamed the doc and added some information on the FTP server, and also linked it from the README.

sbesson:
Thanks.

PR #308, opened 2021-01-21
Bump omero-cli-render to version 0.7.0

sbesson:
This should include support for setting min/max values using the rendering plugin in future deployments of the IDR instance

/cc @pwalczysko @will-moore


manics:
Might as well update OMERO.py to 5.9.0 too?

joshmoore (approved):
LGTM

joshmoore:
(though too bad dependabot doesn't do it ;) )

PR #309, opened 2021-01-25
Update playbooks with Ansible 2.8/Molecule 2.22

sbesson:
--depends-on https://github.com/ome/action-ome-ansible-molecule/pull/1

Not in scope for `prod92` but to be tested when spinning up `prod93`


sbesson:
Used for the deployment on `prod93`

PR #311, opened 2021-01-26
Remove k8s

manics:
We're going to use EBI's K8s cluster instead. In any case these playbooks are probably too old to reliably run.


sbesson:
Conceptually, this clean up matches the decision to stop the maintenance of the VAE as it is currently deployed and look into various alternatives. For the usages of this deployment repository pertaining to the creation of pilot and production IDR instances, this removal should have no incidence. I will test this as part of the `prod93` deployment.

> In any case, these playbooks are probably too old to reliably run.

Does this mean re-executing these playbooks as they stand against the current VAE servers would likely fail?  

More generally, removing these playbooks would raise the question of having some form of actionably contingency/migration plan. In the best scenario, this would happen when we actively work on the migration. In the worst scenario, this would respond to e.g. a failure of the VAE with no easy workaround to bring it back online. /cc @jburel @joshmoore

sbesson (approved):
We'll use next Monday meeting to confirm that everyone in the IDR team is aware of the implications and that the next steps in terms of the VAE migration are captured.

PR #314, opened 2021-02-11
Bump Ansible roles to their latest released version

sbesson:
This PR performs a routine update of the Ansible roles to consume the latest releases. Most of the the increments are patch releases following the Ansible 2.8 upgrade. Below are listed the minor/major bumps with a list of the features 

- ome.docker 3.0.2 -> 3.1.1: allow to override yum repo
- ome.ice 4.0.1 -> 4.2.0: do not use runtime package, use set become
- ome.nfs_mount 1.2.1 -> 1.3.0: support mount state and multiple nfs versions
- ome.nginx 2.0.1 -> 2.1.1: allow to manage nginx without systemd
- ome.nginx_proxy 1.13.0 -> 1.15.1: additional directives, allow to not restart nginx via systemd
- ome.omero_logmonitor 2.1.1 -> 3.0.1: upgrade to Python 3
- ome.omero_server 3.2.0 -> 3.3.1: use ome.python3_virtualenv
- ome.omero_web 3.0.1 -> 3.1.1: use ome.python3_virtualenv
- ome.prometheus 0.4.0 -> 0.5.1: support deployment outside cloud
- ome.prometheus_jmx 0.2.2 -> 0.3.1: prometheus_jmx upgrade

Proposing for the deployment of `prod94`

### DIFF
--- a/ansible/idr-downloads.yml
+++ b/ansible/idr-downloads.yml
@@ -18,7 +18,7 @@
       comment: /OMERO directory from the IDR
     - name: sql
       path: /srv/omero-sql
-      comment: PostgreSQL 9.4 database dump of the IDR
+      comment: PostgreSQL 11 database dump of the IDR
 
   tasks:
 
--- a/ansible/idr-downloads.yml
+++ b/ansible/idr-downloads.yml
@@ -18,7 +18,7 @@
       comment: /OMERO directory from the IDR
     - name: sql
       path: /srv/omero-sql
-      comment: PostgreSQL 11 database dump of the IDR
+      comment: "PostgreSQL {{ postgresql_version }} database dump of the IDR"
 
   tasks:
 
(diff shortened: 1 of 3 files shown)


sbesson:
A few deprecation warnings were logged during the execution of the deployment playbook. Issues have been opened against the affected roles

ansible/idr-downloads.yml
```diff
@@ -18,7 +18,7 @@
       comment: /OMERO directory from the IDR
     - name: sql
       path: /srv/omero-sql
-      comment: PostgreSQL 9.4 database dump of the IDR
+      comment: PostgreSQL 11 database dump of the IDR
```
joshmoore, line 21: :+1: for catching this but perhaps harder to keep up to date than is necessary?

joshmoore (approved):
Generally makes sense. I assume the biggest potential issue could be the prometheus upgrade.

ansible/idr-downloads.yml
sbesson: True, I should be able to use https://github.com/IDR/deployment/blob/fdf2096e7ea4681d1b48c2587b5cf2bf67500d91/ansible/group_vars/omero-hosts.yml#L37 to reduce the maintenance

joshmoore: :+1:

PR #315, opened 2021-02-17
Bump static website

sbesson:


sbesson:
Deployed on `prod93b`

PR #316, opened 2021-02-22
Use IDR OpenStack roles from Ansible Galaxy

sbesson:
All the OpenStack related roles have been now released to https://galaxy.ansible.com/idr with a new compliant name using `_` rather than `-`.

I think Ansible Galaxy is largely case-insensitive so either `idr.*` or `IDR.*` should work.


sbesson:
Used  to deploy `pilot-idr0105`

sbesson:
This has been extensively tested for spinning up multiple pilots which tests all the roles uploaded to Galaxy (with their fixes). I will merge tomorrow if no objection and this can be further tested during the deployment of `prod95`

PR #318, opened 2021-03-05
Bump IDR Bio-Formats to 0.6.6 including the CV7000 fix

sbesson:


ansible/group_vars/omero-hosts.yml
```diff
@@ -66,7 +66,7 @@ omero_server_systemd_limit_nofile: 16384
 
 omero_server_python_addons:
 - omero-cli-render==0.7.0
-- omero-metadata==0.7.0
+- omero-metadata==0.8.0
```
sbesson, line 69: As reported by @dominikl, the API addition introduced in `0.8.0` breaks the metadata population commands in other contexts than the bulk annotation one e.g. the `bulk2map` context. This will need to be fixed in a patch release.

PR #319, opened 2021-03-07
Use openstacksdk rather than shade for resources deletion

sbesson:
Fixes an issue encountered during the deletion of `prod93b` resources related to the detected ports. To reproduce the problem, run a dry-run deletion against an existing environment (or a freshly created one):

```
./scripts/os-idr-delete.py prod95 --type all
```

After confirm the environment to be deleted (`prod95`), the output of the command should list 4 ports to be deleted in the `Networks` section instead of one. 

Debugging reveals that the issue comes from https://github.com/IDR/deployment/blob/e8db873933dbb6e38439b19ba52731460eaeb188/scripts/os-idr-delete.py#L94-L96 which returns all ports and is ignoring the  `device_owner` and `network_id` filters.
While trying to track the issue, I noticed that the final release of `shade`, 1.33.0, happened a year ago with the package is discontinued in favor of `openstacksdk`. Switching the library also fixes the network detection issue. 

With this PR, the same dr=yun command as above only lists a single `network:router_interface` port associated with the environment subnet. 

The swap IP script is also updated to use `openstacksdk`. The only place still dependent on `shade` is the `openstack-idr.py` script used by Ansible. I would propose to get this captured and investigated as a separate issue.

### DIFF
--- a/scripts/os-idr-delete.py
+++ b/scripts/os-idr-delete.py
@@ -3,7 +3,7 @@
 
 import argparse
 from builtins import input
-import shade
+import openstack
 import sys
 
 
@@ -28,8 +28,8 @@ def is_in_idrenv(idrenv, obj):
 
 
 class DeleteResource(object):
-    def __init__(self, cloud, wait):
-        self.cloud = cloud
+    def __init__(self, conn, wait):
+        self.conn = conn
         self.wait = wait
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


joshmoore (commented):
General :+1: for moving to the non-deprecated library. From my POV, it would have been fine to leave `cloud`, but that's more a non-issue.

sbesson:
Yes, I hesitated to keep the diff minimal vs updating the internal variable to match the class name of the API. I also don't feel strongly either way. It's easy enough to update and retest this PR

joshmoore:
> It's easy enough to update and retest this

Don't worry about it.

PR #324, opened 2021-03-18
Add mapr configuration for Protein Supplementary annotations

sbesson:
In the context of the upcoming `idr0072`, this defines a new Protein Supplementary map annotation.


sbesson:
Deployed on `pilot-idr0072` 

Screenshot 2021-03-18 at 21 39 56

sbesson:
Validate by @francesw

PR #325, opened 2021-04-01
Consume latest released ome.omero_server and ome.omero_web roles

sbesson:
Also remove deprecated ome.omego and ome.omero_python_deps roles


sbesson:
Now deployed as `prod96`

PR #328, opened 2021-04-28
Deploy latest version of IDR website

sbesson:
Includes stats for `idr0063` and `idr0072`


sbesson:
Deployed on `prod96` for release

Issue #327, opened 2021-04-27
OMERO.server 5.6.3 upgrade


sbesson:
An initial attempt to upgrade the OMERO.server beyond 5.6.0 had been attempted in https://github.com/IDR/deployment/pull/229 but aborted and captured as https://github.com/ome/omero-server/issues/93. At minimum, the server changes invalidated the Bio-Formats cache files which required

As the latest OpenJDK releases drop support for TLS 1.0/1.1, imports will be broken in the next rolling deployment of IDR (pilot or production). The current CentOS 7 workaround is to set-up self-signed certificates server-side using `omero-certificates` - see https://github.com/ome/ansible-role-omero-server/issues/57. 

In terms of deployment changes, the relevant line is:

```diff
diff --git a/ansible/group_vars/omero-hosts.yml b/ansible/group_vars/omero-hosts.yml
index f1e6763..dade77c 100644
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -64,6 +64,8 @@ omero_server_datadir_bioformatscache: /data/BioFormatsCache
 
 omero_server_systemd_limit_nofile: 16384
 
+omero_server_selfsigned_certificates: True
+
 omero_server_python_addons:
 - omero-cli-render==0.7.0
 - omero-metadata==0.7.0

```

Testing this in a pilot environment with Java upgraded (via `yum update`) results in

```
(base) [sbesson@pilot-idr0072-omeroreadwrite ~]$  /opt/omero/server/OMERO.server/bin/omero import test.fake 
Using session for demo@localhost:4064. Idle timeout: 10 min. Current group: Public
2021-04-27 10:06:27,638 270        [      main] INFO          ome.formats.importer.ImportConfig - OMERO Version: 5.5.5
2021-04-27 10:06:27,654 286        [      main] INFO          ome.formats.importer.ImportConfig - Bioformats version: 0.6.6 revision: 55bcf78bf629ab5dfc5f3e4f677043f11cbc4a57 date: 5 March 2021
2021-04-27 10:06:27,708 340        [      main] INFO   formats.importer.cli.CommandLineImporter - Log levels -- Bio-Formats: ERROR OMERO.importer: INFO
2021-04-27 10:06:28,093 725        [      main] INFO      ome.formats.importer.ImportCandidates - Depth: 4 Metadata Level: MINIMUM
[...]

PR #329, opened 2021-04-29
OMERO.server: upgrade and self-signed certificates

sbesson:
Fixes https://github.com/IDR/deployment/issues/327

- [x] deploy upgrade on `prod97`
- [x] remove the previous cache directory
- [ ] regenerate the memo files for a few studies with many files: idr0044 (original issue), idr0065, idr0094 
- [ ] collect access metrics
- [ ] upgrade the DB with the script of https://github.com/ome/openmicroscopy/pull/6275/files and retest
- [ ] regenerate memo files for all images in IDR

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -6,19 +6,17 @@
 ######################################################################
 # Shared variables
 
-# Force a downgrade from 5.6.1
-# https://github.com/ome/omero-server/issues/93
-idr_omero_server_release: 5.6.0
+idr_omero_server_release: 5.6.3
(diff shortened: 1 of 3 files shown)


sbesson:
Unfortunately, it looks like the server upgrade is blocked on the regeneration of Bio-Formats cache files for datasets which are using absolute paths outside the ManagedRepository (primary `.screen` and `.pattern` based fileset).

The discussion can continue on https://github.com/IDR/deployment/issues/327 regarding the best way to support this scenario and upgrade the server.  Immediately I'll be closing this and opening a new PR with the commits setting up the self-signed certificates.

PR #332, opened 2021-05-07
Install omero-figure

sbesson:
Deploy omero-figure on an IDR environement.

Due to the read-only nature, figures cannot be saved back onto the server.


sbesson:
This is now deployed on` prod98`.

PR #333, opened 2021-05-20
Redirect fluentd logs to idr-logs channel

sbesson:
The goal of this change is to restrict the notify channel to Prometheus (and
possible Munin) alerts and redirect the logs from the various data sources
collected by fluentd (nginx, omero.master) to the logs channel.

### DIFF
--- a/ansible/management-fluentd.yml
+++ b/ansible/management-fluentd.yml
@@ -122,7 +122,7 @@
   vars:
     fluentd_shared_key: "{{ idr_secret_fluentd_shared_key | default('fluentd') }}"
     fluentd_slack_token: "{{ idr_secret_management_slack_token | default(None) }}"
-    fluentd_slack_channel: "idr-notify-{{ idr_environment | default('idr') }}"
+    fluentd_slack_channel: "idr-logs-{{ idr_environment | default('idr') }}"
     fluentd_elasticsearch_host: elasticsearch
     fluentd_uid: "1010"
     elasticsearch_docker_image: "docker.elastic.co/elasticsearch/elasticsearch-oss:6.1.1"


sbesson:
Deployed on `prod97` and manually simulating an outage by stopping the `omero-web` service followed by the `omero-server` service.

The notify channel now only contains the Prometheus alerts while all the logs are collected into the logs channel

Screenshot 2021-05-25 at 10 08 02

I think this is ready for review. Once agreed I would propose for deployment on all the running IDR instances (production and pilots) to simplify the detection of outages.

joshmoore (approved):
What's in `notify` looks good. I can't say that there's nothing in `logs` that I wouldn't _also_ want in `notify`, but I **do** know that when there's too much in `notify` that I lose track so :+1:

sbesson:
Thanks. I had very similar thoughts but I would indeed rather start with limiting the size of `idr-notify` to the critical alerts.

I just came across https://docs.fluentd.org/monitoring-fluentd/monitoring-prometheus which might be something to investigate may we want to create alerts for some Fluentd events.

joshmoore:
Is our top request "keep complaining when it's down"?

sbesson:
Given the prior issues with `prod96`, I would put it at the top of the list. Addressing the `keep` part, as per  https://prometheus.io/docs/alerting/latest/configuration/ I think the alerts should repeat every 4h. I  will test this before merging.

In terms of deployment, since `prod97` is due next week, I will skip `prod96b` but I would propose to apply this to the `pilot-management` and `idr-ftp`

sbesson:
Confirmed that the default alerts are repeated every 3h

Screenshot 2021-05-25 at 22 02 22

joshmoore:
Nice. So you assume that was happening before, and it was just lost in the noise.

sbesson:
> Nice. So you assume that was happening before, and it was just lost in the noise.

I assume in many past cases, we managed to be reactive enough to fix them within 3h. But otherwise, that's my suspicion.

joshmoore:
If you happen to have seen the "3h" setting, I could imagining lowering it.

sbesson:
```git
diff --git a/ansible/management-prometheus.yml b/ansible/management-prometheus.yml
index f68be03..5212ed0 100644
--- a/ansible/management-prometheus.yml
+++ b/ansible/management-prometheus.yml
@@ -73,6 +73,7 @@
   - role: ome.prometheus
     prometheus_docker_network: monitoring
 
+    prometheus_alert_repeat_interval: 5m
     prometheus_alertmanager_slack_webhook: "{{ idr_secret_management_slack_webhook }}"
     prometheus_alertmanager_slack_channel: "#idr-notify-{{ idr_environment | default('idr') }}"
     # The defaults in the prometheus role are (30s 5m 3h)
```

seems to be sufficient to creates alerts every 10min (5min of interval + 5min from `prometheus_alert_group_interval`?)

Screenshot 2021-05-27 at 22 08 04

Which interval would we like to see here?

joshmoore:
If the server is really down? (and I have no evidence that this has triggered when that's not the case) then 30m perhaps? Certainly every hour. Or what would you think?

PR #334, opened 2021-06-01
Reduce interval between successive groups of alters to 25m

sbesson:
If OMERO services are down, this should generate alerts every 30 min

See discussion at https://github.com/IDR/deployment/pull/333. Proposing to deploy this on `prod98`


sbesson:
Deployed on `prod98`, this should now generate an alert every 30min if any system is down

Screenshot 2021-06-02 at 13 13 10

joshmoore (approved):
:heart:

PR #335, opened 2021-06-14
Remove Munin in favor of Prometheus

sbesson:
As discussed at today's IDR meeting, this proposes to remove Munin which was the first monitoring tool introduced in the IDR deployment in favor of Prometheus which has been introduced in https://github.com/IDR/deployment/pull/29.

Target deployment: `prod99`


jburel:
munin has not been in use for a while so it makes sense to remove it
The repository https://github.com/ome/ansible-role-munin should probably be archived in that case

sbesson:
Thanks Merging for deploying, Yes I will archive the source repositories for `ome.munin` and `ome.munin_node` as this was the last place where we were consuming them

PR #336, opened 2021-06-29
Update IDR Bio-Formats version to 0.6.7

sbesson:
Includes the fix for `idr0110`

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -38,13 +38,13 @@ omero_client_jars:
 idr_bf_jars:
   - name: formats-api
     group: idr
-    version: "0.6.6"
+    version: "0.6.7"
   - name: formats-bsd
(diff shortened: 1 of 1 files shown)


sbesson:
This (and other PRs) are failing GitHub Actions with

```
    RUNNING HANDLER [ome.nginx_proxy : restart nginx] ******************************
    fatal: [idr-proxy-docker]: FAILED! => {"changed": false, "msg": "Unable to restart service nginx: Job for nginx.service failed because the control process exited with error code. See \"systemctl status nginx.service\" and \"journalctl -xe\" for details.\n"}
```

which might be related to https://github.com/ome/ansible-role-nginx/pull/11. That's not something I will be able to investigate further until the week of July 12th likely

sbesson:
Successfully tested in `pilot-idr0110` by @dominikl. Merging and deploying onto `prod99`

PR #338, opened 2021-07-15
Embassy Cloud v4: use new Openstack roles with the NFS security group

sbesson:
The new security group is necessary to access NFS data on the Embassy Cloud v4 infrastructure


joshmoore:
:+1:

PR #340, opened 2021-07-27
Use standard 21 port for new FTP deployment

sbesson:
- define a single FTP security group
- add the 21 port to the FTP security group

### DIFF
--- a/ansible/openstack-create-ftp.yml
+++ b/ansible/openstack-create-ftp.yml
@@ -16,7 +16,7 @@
   - name: IDR FTP external access security group
     os_security_group:
       description: External access to IDR FTP (managed by Ansible)
-      name: "{{ idr_environment_idr }}-ftp-external"
+      name: idr-ftp-external
       state: present
 
   - name: IDR FTP external access security group rules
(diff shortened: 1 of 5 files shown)


ansible/openstack-create-ftp.yml
```diff
@@ -16,7 +16,7 @@
   - name: IDR FTP external access security group
     os_security_group:
       description: External access to IDR FTP (managed by Ansible)
-      name: "{{ idr_environment_idr }}-ftp-external"
```
joshmoore, line 19: Can you explain this change?

joshmoore (approved):
One question, but generally looks fine.

ansible/openstack-create-ftp.yml
sbesson: I think the goal of the existing implementation was to be able to have multiple FTP servers e.g. production and staging for testing and test security group changes without affecting production.

This change is effectively mirroring the naming of the other IDR security groups defined in https://github.com/IDR/ansible-role-openstack-idr-security-groups/blob/master/tasks/main.yml . This assumes the settings should be considered as stable API. A potential next step if we are happy with this approach would be to migrate the security group creation upstream to the Ansible role.

joshmoore: :+1:

PR #341, opened 2021-08-05
Increase pilot dev server volume size to 5000G

sbesson:
This should allow to process some of the largest individual volumes published in IDR


joshmoore (approved):
If this ends up being too much of our available space, we can always rollback. :+1:

PR #342, opened 2021-08-06
Override idr_vm_security_groups in pilot dev server provisioning plabook

sbesson:
This should allow the dev server need to have access to the storage for conversion purposes


joshmoore (approved):
:+1:

PR #343, opened 2021-08-11
Zarr server: add service playbook

sbesson:
- create new `idr-export` service playbook to populate `{{ idr_environment_idr }}-dev` server with conversion prerequisites


joshmoore (approved):
:+1:

PR #344, opened 2021-08-18
Add first version of instructions for the IDR object store

sbesson:
- explain bucket creation, policy and CORS configuration setting


sbesson:
I'd propose to get this merged so that we can make step-wise improvements. /cc @joshmoore

PR #346, opened 2021-08-23
Bump omero-web 5.9.2, omero-metadata 0.9.0

will-moore:
Recent releases...


sbesson (approved):
Thanks @will-moore. Deployed on `prod100` and the mapr issue reported this morning in production IDR should now be fixed in the upcoming release.

PR #347, opened 2021-09-10
Update the pilot dev server provisioning playbook to specify the volume type

sbesson:
Consumes the parameter introduced in https://github.com/ome/ansible-role-openstack-volume-storage/pull/3


joshmoore (approved):
LGTM

PR #350, opened 2021-09-23
Migrate the transfer utilities into a separate service playbook

sbesson:
Includes both the Aspera CLI wrapper script as well as rclone

The main driver of this change was to have the Aspera transfer utility deployed on the pilot Zarr (with SSD) as well as the FTP server as secondary data will be generated that will need to be transferred either on NFS or on the S3 storage.


joshmoore (approved):
LGTM

PR #351, opened 2021-09-29
Disable downloads via rsync

sbesson:
--depends-on https://github.com/IDR/ansible-role-openstack-idr-security-groups/pull/5

Port 873 has been historically open on IDR to allow the download of the DB/OMERO data directory via rsync. This has been recently reported as a potential vulnerability.
These changes disable the rsync IDR service. In the absence of logs, it is unclear whether this largely undocumented service was being used or not. Several options were discussed to replace it (Aspera) but we might need to wait until someone comes with a concrete driver/use case to start designing a replacement.


sbesson:
Deployed on `prod102`

PR #354, opened 2021-12-02
Prod103

sbesson:
A few changes for the upcoming `prod103` releases

- deploy the new version of `omero-upload` allowing to in-place upload file annotations
- remove the idr0068 redirect as this study will be published


sbesson:
Deployed on `prod103`

PR #356, opened 2022-01-10
Openstack: handle new volume creation logic

sbesson:
--depends-on https://github.com/ome/ansible-role-openstack-volume-storage/pull/4

Following recent upstream changes in Embassy Openstack, the cloning of volumes is done via the creation of intermediate snapshots. The lifecycle of these snapshots is tightly coupled to the child volumes and the deletion script is adjusted to delete the volumes before the snapshots.

### DIFF
--- a/scripts/os-idr-delete.py
+++ b/scripts/os-idr-delete.py
@@ -58,7 +58,6 @@ class DeleteVolumes(DeleteResource):
         super(DeleteVolumes, self).__init__(conn, wait)
         self.volumes = [v for v in conn.list_volumes()
                         if is_in_idrenv(idrenv, v)]
-        volume_ids = set(v.id for v in self.volumes)
         self.volume_snapshots = [s for s in conn.list_volume_snapshots()
                                  if is_in_idrenv(idrenv, s)]
 
--- a/scripts/os-idr-delete.py
+++ b/scripts/os-idr-delete.py
@@ -71,10 +71,10 @@ class DeleteVolumes(DeleteResource):
         )
 
     def __call__(self):
-        for s in self.volume_snapshots:
-            conn.delete_volume_snapshot(s.id, wait=self.wait)
         for v in self.volumes:
             self.conn.delete_volume(v.id, wait=self.wait)
+        for s in self.volume_snapshots:
+            conn.delete_volume_snapshot(s.id, wait=self.wait)
 
 
 class DeleteNetworks(DeleteResource):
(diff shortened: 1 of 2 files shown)


scripts/os-idr-delete.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, conn, idrenv, wait=True):
                         if is_in_idrenv(idrenv, v)]
         volume_ids = set(v.id for v in self.volumes)
         self.volume_snapshots = [s for s in conn.list_volume_snapshots()
-                                 if s.volume_id in volume_ids]
+                                 if is_in_idrenv(idrenv, s)]
```
joshmoore, line 63: Is `volume_ids` still needed?

joshmoore (commented):
One minor question, otherwise :+1:

scripts/os-idr-delete.py
sbesson: I don't _think_ so but let me get through an example to make sure I don't forget anything

Taking the example of the current `prod103` volumes. Moving forward, are two types of volume snapshots that might be created and associated to the source volumes via `volume_id`:

- snapshots created at release time via https://github.com/IDR/deployment/blob/master/scripts/os-idr-snapshot.sh i.e. `prod103-database-db-20211207` ... primarily for backup purposes
- snapshots created when deploying new pilots. As per the latest changes of the upstream role, these snapshots will be prefixed by the pilot environment e.g. `pilot-idr0119-omeroreadwrite-db-2022-01-10T12:57:37Z` and their lifecycle is bound to the lifecyle of the derived volumes

In terms of the deletion workflow:
- when `prod103` is decommissioned, we want only the `prod103-*` volumes and snapshots to be deleted but leave derived pilot snapshots/volumes untouched
- when pilots are decommissioned, we want only the `pilot-idrxxx` volumes and snapshots to be deleted

So I think `volume_ids` is now irrelevant for managing snapshots and we are using the environment prefix like the rest of the resources. If so, I will push another commit to remove the unused variable.
Do you see some scenario I might be missing?

sbesson: Actually while cleaning up `prod103` resources, I see why the code was implemented this way

```
openstack.exceptions.BadRequestException: BadRequestException: 400: Client Error for url: https://uk1.embassy.ebi.ac.uk:8776/v3/ecb43df152dc4e8bbbba2896fac99f51/volumes/44910ed3-c2bb-4bff-971d-789d37f8523b, Invalid volume: Volume status must be available or error or error_restoring or error_extending or error_managing and must not be migrating, attached, belong to a group, have snapshots or be disassociated from snapshots after volume transfer.
```

While the workflow could be viable for pilot deployments (as we do not snapshot them), for production deployment, this is effectively chaining all the production volumes via their snapshots. I'll enquire with the Embassy team.

PR #358, opened 2022-01-19
Handle Openstack volume flattening

sbesson:
--depends-on https://github.com/ome/ansible-role-openstack-volume-storage/pull/5

This PR the logic for creating volumes from existing sources. As this step now involves the flattening of the volumes and requires several hours, the tasks are executed in a separate playbook which should be executed before running the provisioning playbook as usual.

Since the OpenStack volume names are no longer prefixed by the VM name (as this variable is optional), the volume names are also reviewed.  The prefixing of the volume names by `{{ idr_environment_idr }}` is preserved as the various deletion/management script uses this convention.

Finally, the deletion script changes made in https://github.com/IDR/deployment/pull/356 are all reverted since OpenStack requires volume snapshots to be deleted before deleting the source volumes. With the flattening of the volume at creation time, the intermediate volume snapshots should now be deleted as part of the role when the volume is attached leaving only the backup volume snapshots created on release.


sbesson:
Following the execution of the new volume creation playbook

```
(idr-ansible) (base) sbesson@ls30630:idr (master) $ openstack volume show test104-omerodata | grep _at
| created_at                   | 2022-01-19T16:41:32.000000                                       |
| updated_at                   | 2022-01-19T18:36:09.000000                                       |
(idr-ansible) (base) sbesson@ls30630:idr (master) $ openstack volume show test104-db | grep _at
| created_at                   | 2022-01-19T16:41:26.000000                                       |
| updated_at                   | 2022-01-19T18:18:55.000000                                       |
(idr-ansible) (base) sbesson@ls30630:idr (master) $ openstack volume show test104-nginxcache | grep _at
| created_at                   | 2022-01-19T16:41:37.000000                                       |
| updated_at                   | 2022-01-19T16:49:05.000000    
```

i.e. ~2h to clone the three volumes from a previous production environment.

I will continue tomorrow with the rest of deployment and test the deletion and recreation of the environment.

sbesson:
Merging and I'll use this to create `prod105`

PR #359, opened 2022-01-25
Deploy search engine

khaledk2:
I have added playbooks to deploy searchengine, searchengine client and ELasticsearch.
"management-searchengine.yml" is used to configure and run all three applications.
There is a variables file (searchengine_vars.yml) that the user needs to customize before running the playbook
After deploying the apps using the playbook, it is needed to run another playbook (run_searchengine_index_cache_services.yml) for caching and indexing
As the caching and indexing processes take a long time, there are another two playbooks that enable the user to check if they have finished or not i.e. check_indexing_service.yml and check_caching_service.yml.


ansible/management-searchengine.yml
```diff
@@ -0,0 +1,253 @@
+# Search ngine + Search Engine Client + Elasticsearch
+
+#- hosts: "{{ idr_environment | default('idr') }}-management-hosts"
```
sbesson, line 3: Should this line be uncommented?

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,253 @@
...
+#    - /vars/external_vars.yml
+#- hosts: all
+#  vars_files:
+#    file: searchengine_vars.yml
+
+- name: Deploying search engine
+  connection: local
+  hosts: local
+  vars_files:
+    searchengine_vars.yml
```
sbesson, line 21: Moving the variables under `group_vars` should allow them to be automatically picked up - see https://docs.ansible.com/ansible/2.8/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,253 @@
...
+
+  pre_tasks:
+    - raw: sudo yum update -y
```
sbesson, line 24: At the moment, the first playbook executed after the provisioning (https://github.com/IDR/deployment/blob/master/ansible/idr-00-preinstall.yml#L7) upgrade all the packages before installing the services. So this should be unnecessary and can be removed.

ansible/searchengine_vars.yml
```diff
@@ -0,0 +1,13 @@
+apps_folder: host_data_folder
```
sbesson, line 1: Should that become `/data` initially?

sbesson (changes requested):

ansible/management-searchengine.yml
khaledk2: This line has been uncommented and moved to replace local in  (host: local) and removed (connection: local) as I was testing them locally. I have done the same for the other playbooks. I think this should work fine?

ansible/management-searchengine.yml
khaledk2: I have moved it

ansible/management-searchengine.yml
khaledk2: Ok, I have removed it and removed the link from the playbooks.

ansible/searchengine_vars.yml
khaledk2: Yes, I have updated the vars file.

sbesson:
After copying the vars files to match the name of the group and removing the variable pointing to a local path

```diff
(idr-ansible) (base) sbesson@ls30630:ansible ((db001bb...)) $ diff group_vars/searchengine_vars.yml group_vars/management-hosts.yml 
11d10
< ansible_python_interpreter: path/to/bin/python
```

the playbook executed until 

```
TASK [configure elasticsearch  for docker searchengine] **********************************************************************************************************************************************
fatal: [test104-management]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/usr/bin/python"}, "changed": false, "msg": "Docker SDK for Python version is 1.10.6 (test104-management.novalocal's Python /usr/bin/python). Minimum version required is 2.1.0 to set auto_remove option. Try `pip uninstall docker-py` followed by `pip install docker`."}

PLAY RECAP *******************************************************************************************************************************************************************************************
test104-management         : ok=12   changed=11   unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   
```

Possible options to move forward are:
- deploy the `auto_remove` option for now (does any behavior depend on it?)
- review how the Docker SDK for Python is install and upgrade and/or create a local virtual environment with a recent version of the `docker` module

khaledk2:
auto_remove instructs to delete the container after it runs. I think we may comment on it for the time being, what do you think?

sbesson:
Agreed, let's comment it out and come back to it later in the testing.

khaledk2:
Sorry, I should mention before that I have commented on auto_remove and push the playbooks yesterday.

ansible/group_vars/searchengine_vars.yml
```diff
@@ -0,0 +1,13 @@
...
+database_server_url: "{{ omero_db_host_ansible }}"
+database_port: 5432
+database_name: idr
+database_username: omeroreadonly
+database_user_password:  "{{ idr_secret_postgresql_password_ro | default('omero') }}"
+searchenginecache_folder: /data/searchengine/searchengine/cacheddata/
+search_engineelasticsearch_docker_image: docker.elastic.co/elasticsearch/elasticsearch:7.16.2
+searchengine_docker_image: khaledk2/searchengine:0.1
+searchengineclient_docker_image: khaledk2/searchengineclient:0.1
+ansible_python_interpreter: path/to/bin/python
```
sbesson, line 11: With the commenting out of `auto_remove`, this line should be commented

docs/searchengine_deployemnt.md
```diff
@@ -0,0 +1,21 @@
+Searchengine installation and configuration using Ansible:
+==========================================================
+
+There is an ansible playbook (management-searchengine.yml) that has been written to deploy the apps:
+* It will config and run searchengine, Elasticsearch and searchengine client
+* It will configure and create the required folders
+* It will configure the three apps and run them
+* There is a variables file (searchengine_vars.yml) that the user needs to edit before running the playbook
```
sbesson, line 8: This file should be renamed as `groups_vars/management-hosts.yml` until we decide on a better group/infrastructure

ansible/group_vars/searchengine_vars.yml
```diff
@@ -0,0 +1,13 @@
+apps_folder: /data
+database_server_url: "{{ omero_db_host_ansible }}"
```
sbesson, line 2: When executing the playbook, this leads to `'omero_db_host_ansible' is undefined`. I think the metadata will need to be extracted using a `set_fact` task similar to https://github.com/IDR/deployment/blob/db10eafc4c7f131ff14d4cd639a59a94a2751478/ansible/idr-omero.yml#L27-L32

sbesson (changes requested):

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,240 @@
+# Search ngine + Search Engine Client + Elasticsearch
+
```
sbesson, line 2:
```suggestion
- hosts: "{{ idr_environment | default('idr') }}-database-hosts"

```

is a change I had to apply in my local deployment to make the playbook run of `test104` as it allows to load the variables from the database hosts in order to set the `database_server_url` variable

sbesson (changes requested):

ansible/check_caching_service.yml
```diff
@@ -0,0 +1,20 @@
+#Issue setup ip address inside the hpa config file for postgres to accept the connection from it  
+- name: Check caching
+  hosts: "{{ idr_environment | default('idr') }}-management-hosts"
+
+  tasks:  
+  - name: Get infos on container
+    docker_container_info:
```
sbesson, line 7:
```suggestion
    become: yes
    docker_container_info:
```

I had to insert this to avoid a Permission denied

sbesson:
Added a minimal configuration allowing to proxy the 5567 port under the `/searchengine` endpoint

```
TASK [ome.nginx_proxy : nginx | proxy config] *************************************************************************************************
--- before: /etc/nginx/conf.d/proxy-default.conf
+++ after: /Users/sbesson/.ansible/tmp/ansible-local-26677fmuyzkn/tmp_tkpwlmz/nginx-confd-proxy.j2
@@ -253,15 +253,6 @@
 
     }
 
-    location ^~ /searchengine {
-        proxy_pass http://searchengine/;
-        proxy_redirect http://searchengine $scheme://$server_name;
-
-
-        proxy_ignore_headers   "Set-Cookie" "Vary" "Expires";
-        proxy_hide_header Set-Cookie;
-    }
-
 
     add_header Access-Control-Allow-Origin $allow_origin;
 
​
changed: [test104-proxy] => (item={'nginx_proxy_is_default': True, 'nginx_proxy_additional_directives': ['add_header Access-Control-Allow-Origin $allow_origin']})
ok: [test104-proxy] => (item={'nginx_proxy_server_name': 'cachebuster', 'nginx_proxy_listen_http': 0, 'nginx_proxy_ssl': False, 'nginx_proxy_cachebuster_enabled': True, 'nginx_proxy_backends': [{'name': 'omerocached', 'location': '~ /webclient/metadata_*|/webclient/render_*|/webclient/get_thumbnail*|/webgateway/metadata_*|/webgateway/render_*|/webgateway/get_thumbnail*|/webclient/api/*|/webclient/search/*|/api/*|/webclient/img_detail/*|/iviewer/*|/figure/*|/gallery-api/*|/mapr/*', 'server': 'http://omeroreadwrite', 'cache_validity': '1d', 'read_timeout': 900}, {'name': 'omerostatic', 'location': '~ /static/*', 'server': 'http://omeroreadwrite', 'cache_validity': '1d'}, {'name': 'omero', 'location': '/', 'server': 'http://omeroreadwrite'}]})
ok: [test104-proxy] => (item={'nginx_proxy_server_name': 'idr-demo.openmicroscopy.org', 'nginx_proxy_ssl': True, 'nginx_proxy_redirect_map_locations': [], 'nginx_proxy_direct_locations': [{'location': '/', 'redirect301': '$scheme://idr.openmicroscopy.org$request_uri'}], 'nginx_proxy_backends': []})
​
TASK [ome.nginx_proxy : nginx | proxy upstream servers] ***************************************************************************************
--- before: /etc/nginx/conf.d/proxy-upstream.conf
+++ after: /Users/sbesson/.ansible/tmp/ansible-local-26677fmuyzkn/tmplufdxnu2/nginx-confd-proxy-upstream.j2
@@ -13,6 +13,3 @@
 upstream omeroreadwrite {
   server 192.168.3.22;
 }
-upstream searchengine {
-  server 192.168.3.120:5567;
-}
```

Following this morning's discussion, we are currently running into two issues:

- the static files are missing from the endpoint. Khaled is looking into the Nginx configuration
- the indexing process failed with a statement timeout. Khaled identified the issue as memory related and reduced the number of rows processed concurrently to allow the indexing to run. The `management` VM currently used for the deployment is relatively small with 8GB RAM and 4 VCPUs. When moving to a production state, we might consider hardening this configuration and provisioning the searchengine VM with more resources similarly to the OMERO ro/rw VMS.

khaledk2:
- I should have written that before. I have changed one line in the searchengine section in the proxy-default.conf Nginx configuration file to add  searchengine" subdomain to the HOST header
              ` proxy_set_header   Host $host/searchengine`
- Also, I have made some modifications in the search engine client code, to work when its URL is not the domain root.
- I have renamed the "static" folder to the "searchengineclientstatic" folder in the search engine client.
- I have changed the searchengine code to allow the number of rows to be configured externally (in the app configuration) so it can be customized according to the host machine configuration.
- I have changed the deployment files and added "cache_rows" variable in management-searchengine-hosts.yml to set the number of rows and added an additional step to the management-searchengine.yml to set his configuration.
- I have added searchengine_secret_key and searchengineclient_secret_key variables to set  SECRET_KEYs for each searchengine and searchengine client and added steps inside the deployment yml file to configure them.
- The docker images for each of searchengine and searchengineclient are hosted in my Docker Hub account, they are built and pushed manually. I have created GitHub actions to build and push them automatically to the openmicroscopy DockerHub account, they are in the testing stage.

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,299 @@
...
+- name: Deploying search engine
+  hosts: "{{ idr_environment | default('idr') }}-management-hosts"
+
+  tasks:
+  - name: Get database host
+      set_fact:
```
sbesson, line 8: This block needs to be indented the same way as the line above

sbesson (changes requested):

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,298 @@
...
+      volumes:
+      - "{{ apps_folder }}/searchengine/client:/etc/searchengineclient/"
+
+  - name: configure app data folder for search client
+    become: yes
+    docker_container:
+      image: "{{ searchengineclient_docker_image }}"
+      name: searchengineclient_search_uri
+      cleanup: True
+      auto_remove: yes
```
sbesson, line 258: This suffers from the same issue as https://github.com/IDR/deployment/pull/359#issuecomment-1023199966. Can it be safely commented for now?

sbesson (commented):
With the last set of commits, I was able to successfully deploy the search engine application and launch the indexing/caching processes on a fresh `test104` deployment.

The following local changes to the `idr-proxy.yml` playbook/variables were also applied to deploy the client under the `/searchengine` endpoint:

```diff
diff --git a/ansible/group_vars/proxy-hosts.yml b/ansible/group_vars/proxy-hosts.yml
index e082a821..63dc5cd8 100644
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -38,6 +38,8 @@ nginx_proxy_upstream_servers:
   servers: "{{ omero_omeroreadonly_hosts_external | map('regex_replace', '^(.*)$', '\\1:4065') | sort }}"
 - name: omeroreadwrite
   servers: "{{ omero_omeroreadwrite_hosts }}"
+- name: searchengine
+  servers: "{{ searchengine_hosts | map('regex_replace', '^(.*)$', '\\1:5567') | sort }}"
 
 # The regex is getting complicated, so unroll it into a list and join
 _nginx_proxy_omero_locations:
@@ -100,11 +102,19 @@ _nginx_proxy_backends_prometheus_federate:
   server: "http://{{ management_host_ansible | default('localhost') }}:9090/federate"
   cache_validity: 15s
 
+_nginx_proxy_backends_searchengine:
+- name: prometheusfederate
+  location: "^~ /searchengine"
+  server: http://searchengine/
+  host_header: "$host/searchengine"
+
+
 nginx_proxy_backends: >
   {{ _nginx_proxy_backends_omero +
      _nginx_proxy_backends_omerowebsockets +
      _nginx_proxy_backends_grafana_render +
-     _nginx_proxy_backends_prometheus_federate
+     _nginx_proxy_backends_prometheus_federate +
+     _nginx_proxy_backends_searchengine
   }}
 
 
diff --git a/ansible/idr-proxy.yml b/ansible/idr-proxy.yml
index edc0db47..adb601f2 100644
--- a/ansible/idr-proxy.yml
+++ b/ansible/idr-proxy.yml
@@ -61,6 +61,12 @@
             idr_environment | default('idr') + '-management-hosts'][0]]
             ['ansible_' + (idr_net_iface | default('eth0'))]['ipv4']['address']
         }}
+      searchengine_hosts: >-
+        {{
+          groups[idr_environment | default('idr') + '-management-hosts'] |
+          map('extract', hostvars,
+            ['ansible_' + (idr_net_iface | default('eth0')), 'ipv4', 'address']) | list
+        }}
     when: groups[idr_environment | default('idr') + '-management-hosts'] is defined
 
   roles:
```

A few inline comments and the client probably needs some testing once the indexing/caching has completed.
From a code perspective, I think we are approaching thew point where this playbook can be safely merged into the repository. Importantly, as things stand, this app is not included by default and needs to be deployed manually. Probably the biggest question for the IDR team is whether we would consider deploying it on all `prod` deployment as an experimental endpoint and/or the steps to move towards this target.

khaledk2:
I have pushed changes to run on the searchengine-hosts group and removed hdf5 caching service as all the cached data now is saved in  Elasticsearch.

ansible/management-searchengine.yml
```diff
@@ -0,0 +1,298 @@
+# Search ngine + Search Engine Client + Elasticsearch
```
sbesson, line 1: Optionally, this playbook could also be renamed as `idr-searchengine.yml` since it's no longer tied to the management VM

sbesson (changes requested):

khaledk2:
I have renamed the files, and increasd cache_rows to be 50000, I think we can increase it more than that

khaledk2:
I have reverted renaming dockermanager-hosts.yml and renamed the files

sbesson (commented):
With the change to the `hosts` section of `idr-searchengine.yml`, I was able to deploy the searchengine stack on `pilot-idr0000` and start the indexing process which completed in ~12h.

khaledk2:
I have added a variable to the "searchengine-hosts.yml"; its value equals the URL prefix part (searchengineapi)
`searchengineurlprefix:` "searchengineapi"

It will be used to set the script_name when running gunicorn 

Also, I have changed the Nginx configuration at the searchengineapi section

` location ^~ /searchengineapi {
        proxy_pass http://searchengineapi/searchengineapi;
        proxy_redirect http://searchengineapi/searchengineapi $scheme://$server_name;
    } `

sbesson (approved):
With the latest set of changes and #367, I was able to successfully deploy the search engine stack onto a newly create pilot using the new group.

The API is available when forwarding the port 5577 and the client is available when accessing the port 5556.

The playbook is currently set up so that it will only run when executed manually against VMs with the correct groups.

As discussed this morning as part of the weekly IDR call, merging this so that we can make incremental progresses towards a production release of the new service via smaller PRs. I will capture the outstanding issues as todos.

PR #362, opened 2022-03-08
Consume IDR Bio-Formats 0.6.8 in IDR deployment

sbesson:
Includes the OME-TIFF improvements from https://github.com/IDR/bioformats/pull/27


sbesson:
Since both `idr0120` and `idr0133` are currently in-scope for `prod107`, I intend to merge and deploy this PR after releasing `prod106` if no vetoes.

PR #372, opened 2022-06-01
Update playbooks to deploy searchengine in production deployment

### LINKED ISSUE #370: prod107: search engine deployment
The deployment of a first version of the search engine stack is a target of the upcoming `prod107` release. #359 introduces the playbook allowing to deploy the stack while #367 contains the logic to define a new group of servers where the service should be deployed.

While #367 focuses on deploying the service in the simple context of a pilot VM where all the services are colocated on a single nde, for the scope of `prod107` we will need to decide how to deploy it the multi-nodes architecture used for production deployments.

The current set of instances (and their relationship) created for each deployment can be loosely summarized by:

`idr-database` -> `idr-omeroreadonly-1,idr-omeroreadonly-2,idr-omeroreadonly-3,idr-omeroreadonly-4, idr-omeroreadwrite` -> `idr-proxy`
`idr-database,idr-omeroreadonly-1,idr-omeroreadonly-2,idr-omeroreadonly-3,idr-omeroreadonly-4, idr-omeroreadwrite` ->`idr-management`

Listing the various architectures available

## Option 1: deploy the app in the `management` instance

Pros: it benefits from the Docker prerequisites being installed in the `management` instance (Currently used for monitoring), it's the strategy originally used for #359, currently deployed on `test104`
Cons: the compute capacity of the `management` VM is limited esp. for a full indexing, consuming the search endpoint from the `omero` nodes would require to go through the proxy

## Option 2: deploy the app in the `omeroreadwrite` instance

Pros: this is a more scaled version of the option 1, it uses the fact the `omeroreadwrite` server has a larger compute capacity. Additionally, this makes use of the capacity of `omeroreadwrite` which is currently unused when the deployment is moving to production (except for minor DB updates like adding DOIs/publication)
Cons: same as above, consuming the search endpoints from the `omeroreadonly` nodes currently requires to go through the proxy without additional nginx configuration

## Option 3: deploy the app in a new `searchengine` instance

Pros: allows to tailor compute/storage capacity of the instance to the exact needs of the app. Allows the various `omero` instances to access  the `searchengine` service in the same way as the `database` is accessed
Cons: requires  1 more instance would be created per production deployment and probably needs to be reviewed with the global tenancy capacity

## Option 4: deploy the app across all `omero` instances

Pros: for indexing, this would keep the benefit of option 2 and use the compute capacity of `omeroreadwrite`, if we are thinking of integrating with omero-web or idr-gallery, it colocates the service and simplifies. Also this starts scaling the service in the same way as the OMERO.web servers
Cons: probably requires additional thoughts on how to distribute the data especially the elasticsearch database, possible moving towards an ElasticSearch cluster (4a) or moving ElasticSearch to yet another instance (option 4b)

sbesson:
Fixes #370

As discussed on the issue, this PR implements option 3 and creates a new VM of similar compute capacity to `omeroreadwrite` and some dedicated storage.

To be deployed on `test107`

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -100,11 +100,18 @@ _nginx_proxy_backends_prometheus_federate:
   server: "http://{{ management_host_ansible | default('localhost') }}:9090/federate"
   cache_validity: 15s
 
+_nginx_proxy_backends_searchengineapi:
+- name: searchengineapi
+  location: "^~ /searchengineapi"
+  server: "http://{{ searchengine_host_ansible | default('localhost') }}:5577/searchengineapi"
+  host_header: "$host/searchengine"
+
 nginx_proxy_backends: >
   {{ _nginx_proxy_backends_omero +
      _nginx_proxy_backends_omerowebsockets +
      _nginx_proxy_backends_grafana_render +
-     _nginx_proxy_backends_prometheus_federate
+     _nginx_proxy_backends_prometheus_federate +
+     _nginx_proxy_backends_searchengineapi
   }}
 
 
--- a/ansible/idr-02-services.yml
+++ b/ansible/idr-02-services.yml
@@ -9,3 +9,6 @@
 
 ### Transfer services
 - include: idr-transfer.yml
+
+## Search services
+- include: idr-searchengine.yml
--- a/ansible/idr-proxy.yml
+++ b/ansible/idr-proxy.yml
@@ -8,9 +8,13 @@
     {{ idr_environment | default('idr') }}-omero-hosts
 
 
-# Load hostvars (managment server)
+# Load hostvars (management server)
 - hosts: >
     {{ idr_environment | default('idr') }}-management-hosts
+
+# Load hostvars (search engine server)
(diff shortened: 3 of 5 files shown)


sbesson:
The volumes of `prod107` are now ready for redeployment. I would propose to get this merged so that we redeploy it with the provisionin infrastructure decided for the searchengine backend. Improvements and to the app/deployment can then be opened as follow-up PRs. @khaledk2 would this work from your side?

sbesson:
All modifications would happen in the `idr-searchengine.yml` playbook, is that correct? IF so, I'd propose to get this in and let you open a follow-up PR cleaning the deployment of the initial client?

khaledk2:
Some modifications are needed also in `searchengine-hosts.yml`

sbesson:
Thanks @khaledk2 and fair enough. As these variables are only used in the post-provisioning steps (and specifically during the execution of the [idr-02-services.yml](https://github.com/IDR/deployment/blob/ef48731726acd32298c034f94c1f64653188cd86/ansible/idr-02-services.yml) playbook, this should still be fine to handle separately. I'll merge this and leave you to open the cleanup PR

PR #374, opened 2022-06-07
Remove all the references to the search engine client deployment

khaledk2:
This PR removes all the references to the search engine client deployment as it has been added as a web app to IDR deployment so there is no point in including it.


sbesson (approved):
The cleanup makes sense. Deployed on `prod107` and I can access e.g. the `/searchengineapi/apidocs` endpoint. Anything specific to test?

khaledk2:
No, this should be fine. 
I have tested the API endpoints using the Swagger documents ([`/searchengineapi/apidocs`](http://idr-testing.openmicroscopy.org/searchengineapi/apidocs/)) and it looks fine.

PR #378, opened 2022-06-21
add task to configure  IDR_TEST_FILE_URL

khaledk2:
Although this configuration item has a default value in the app configuration file, we may need to refer to another file  for verification rather than the default one `https://raw.githubusercontent.com/IDR/idr.openmicroscopy.org/master/_data/studies.tsv `


sbesson:
Merging for the redeployment of `prod107`. 
This configuration is a good illustration of a more general concern that the `omero_search_engine` code base currently includes contains many tasks specific to IDR. I suspect more discussions will be needed on the scope of this service. Depending on the decsisions we take collectively, we might need to come back to the architecture/deployment and for example extract IDR-specific steps. Immediately the priority is to get something available publicly so that we can build experience and decide on the following steps.

PR #379, opened 2022-06-24
Configure self-signed certificates on all OMERO hosts

sbesson:
The self-signed SSL certificates were initially deployed on the read-write hosts to fix import issues but they might be required on all read-only servers to fix access via the Java API

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -110,6 +110,7 @@ omero_server_config_set:
   # Disable all components except Blitz and Tables
   omero.server.nodedescriptors: "master:Blitz-0,Tables-0"
 
+omero_server_selfsigned_certificates: True
 
 ######################################################################
 # Other dependent role vars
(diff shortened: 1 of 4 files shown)


sbesson:
@dominikl these changes are now deployed on `test107` and should generate and configure self-signed SSL certs under `/data/OMERO/certs` on each read-only VM. Could you give it a try to see if it fixes the `SSLHandshakeException` when accessing the testing resource via the Java API?

I need to spend more time assessing the impact of https://github.com/IDR/deployment/pull/379/commits/eaa20a22d03653f5c795dd6ec6c54d78bec2de00 before moving this to production, possibly recreating a complete `test107` deployment to confirm this does not cause issue on initial deployments.

dominikl:
👍 I can confirm, can connect now to idr-testing with the java gateway without problem!

dominikl (approved):

sbesson:
Successfully retested on a fresh redeployment of `test107`. Merging and updating the read-only deployments of `prod107`

sbesson:
Now deployed on `prod107`

PR #382, opened 2022-08-15
bump idr-gallery to 3.6.2

will-moore:
In order to deploy latest change in https://github.com/IDR/idr-gallery/pull/10


sbesson (approved):
Successfully deployed on `prod109`

PR #385, opened 2022-09-29
Replace deprecated include directive by import_playbook

sbesson:
Updates the IDR playbooks as these directives will be dropped in Ansible 2.12


sbesson:
This has been successfully used to deploy `pilot-idr0125`. Next step will be to run it in against full production deployment context but there is only a handful set of steps which have been untested so I'll just get this merged and I will capture potential issues as follow-ups.

PR #391, opened 2023-01-30
change searchEngine version to 0.5.0

khaledk2:
I have updated `searchengine-hosts.yml` with the new searchengine version.

PR #392, opened 2023-01-31
Run searchengine container using a septate playbook

khaledk2:
This PR is used to run the searchengine container using a separate playbook and can be used to update the searchengine when it is needed. 
This comes from this discussion https://github.com/IDR/deployment/pull/391#issuecomment-1410024671.


ansible/idr-02-services.yml
```diff
@@ -12,3 +12,4 @@
 
 ## Search services
 - import_playbook: idr-searchengine.yml
+- import_playbook: idr-run-searchengine-container.yml
```
sbesson, line 15: Without reading the content, I find these names make it hard to understand the intent of each playbook. See below for a proposal

ansible/idr-searchengine.yml
```diff
@@ -1,4 +1,4 @@
-# Search ngine +  Elasticsearch
+# Configure Search engine +  run Elasticsearch nodes
```
sbesson, line 1: As per https://github.com/IDR/deployment/pull/391#issuecomment-1410024671, I understand our goal is to redeploy the search engine service without changing elastic search. To achiev this, would it make sense to have the first playbook focusing solely on elasticsearch and the second on searchengine?
Practically, this would mean migrating all tasks configuring  `{{ apps_folder }}/searchengine/searchengine` to the second playbook.

khaledk2:
It is a good idea!
I have modified it so there is a separate playbook to run and configure each of elasticsearch (`idr-elasticsearch.yml`) and searchengine (`idr-searchengine.yml`).

ansible/run_searchengine_index_service.yml
```diff
@@ -1,23 +1,157 @@
-#Issue setup ip address inside the hpa config file for postgres to accept the connection from it
-- name: Deploying search engine cache and indexing
```
sbesson, line 2: Is this playbook effectively a copy of `idr-searchengine.yml`? What are the differences?

ansible/idr-elasticsearch.yml
```diff
@@ -0,0 +1,164 @@
+# configure and run Elasticsearch for the searchengine
+
+- hosts: "{{ idr_environment | default('idr') }}-database-hosts"
```
sbesson, line 3: Loading the host variables from the database VM was used in the searchengine playbook for setting the `database_server_url` variable.
Is it correct that it is not required at all for setting up Elasticsearch? In this case, this line can simply be removed

sbesson (commented):
Two additional comments. Otherwise, I think the separation of the services makes sense and I would propose to test it in a full deployment (`test115` or `prod115`) next.

ansible/run_searchengine_index_service.yml
khaledk2: Sorry, I have overwritten this file, I have restored its original contents. It is used for indexing.
This leads me to update the searchengine documentation `searchengine_deployment.md`

ansible/idr-elasticsearch.yml
khaledk2: I have removed it from this playbook.

sbesson:
Thanks @khaledk2, the latest state of the changes look good from a code review perspective. I'll integrate them in the next full deployment.

khaledk2:
I have updated the searchengine version to 0.5.1.  I think it may be a good idea to update the searchengine manually in prod114 as we do not have time if something goes wrong.

We need first to stop the container

`sudo docker stop searchengine`

then we need to run the new version using this command:

`sudo docker run -d  --name searchengine  --rm  -v /data/searchengine/searchengine/:/etc/searchengine/ --network=searchengine-net  -p 5577:5577  openmicroscopy/omero-searchengine:0.5.1  run_app searchengine`

I can update it if you do not  mind.

sbesson:
@khaledk2 go for it and agreed on being cautious for the upcoming release. I'll test this PR with `prod115`

khaledk2:
`searchengine `version `0.5.1` has been deployed in `prod114`.

sbesson (approved):
Successfully deployed on `pilot-idr000`. We might want to run a minimal indexing/search test to confirm the functionality is identical to `prod114`. Overwise, happy for this to be included in `prod115`

PR #395, opened 2023-02-06
bump ome.omero_server to 4.1.0

jburel:
bump ome.omero_server to 4.1.0


sbesson (approved):
Successfully deployed on `pilot-idr0000` and the tables service remain functional

PR #397, opened 2023-03-15
Molecule 3 upgrade

sbesson:
CI builds have been broken for the last few weeks with failures appearing during the early phases of the Molecule test. After some minimal investigation, I did not find a simple way to fix the issue using the existing environment.

This PR proposes to address the failures with the following changes:
- the Molecule configuration is upgraded to comply with Molecule 3
- the upgrade of `ca-certificates` (mandatory to install the PSQL repositories using the base CentOS 7 image) is moved to the prepare playbook
- the GitHub workflow is updated to install the current deployment prerequisites (Python 3.8+, Ansible 3.10) as well as Molecule 3.x and the new `molecule-plugins[docker]` package
- the `idempotence` stage fails the `publicidr` scenario and is temporarily ignored. The issue can be tracked down to the `ome.postgresql` role and will be tackled separately

This PR raises a few questions beyond the scope of this repository and relevant for the other OME Ansible roles/playbooks repositories where a similar upgrade is probably in scope:
- should we keep maintaining the https://github.com/ome/ome-ansible-molecule meta-package and propose an upgrade of Molecule to v3 or specify the dependencies individually?
- should we. keep using an action wrapping `molecule test`? As shown below the lines difference is minimal and the usage of another layer makes the testing of these types of changes significantly more complex


joshmoore (approved):
* the Molecule configuration is upgraded to comply with Molecule 3 :+1:
* the upgrade of ca-certificates ... is moved to the prepare playbook :+1:
* the GitHub workflow is updated to install the current deployment prerequisites :+1:
* the idempotence stage fails the publicidr scenario and is temporarily ignored :+1:

* should we keep maintaining the https://github.com/ome/ome-ansible-molecule ❓ The changes here look to be fairly straight-forward. I'm not strongly tied to the meta package.
* keep using `molecule test` :question: Again no string feelings. Being able to look in the workflow and see what's happening might be a bonus. If there are a large N of these files though it could get annoying if small changes are needed.

sbesson:
Thanks for the review @joshmoore. If we all agree, I will probably go ahead and open a PR to deprecate the ome-ansible-molecule meta-package and we can use the strategy proposed here to upgrade Ansible playbooks and roles workflows as needed.

As a bonus, the last commit upgrades the Ansible version to the current stable 7.x version. The fact that the tests kept passing is a positive sign and matches @pwalczysko's suggestion in https://github.com/ome/omero-deployment-examples/pull/70 that our infrastructure is largely compatible with recent versions of ansible-core/ansible.

If no objection, I'll merge this and probably test a real pilot deployment using Ansible 7.x.

PR #398, opened 2023-03-21
Archive searchengine logs

khaledk2:
The searchengine log contains important information and we would like to keep it after decommissioning the instance. So, I have modified `archive-logs.yml` to archive the searchengigs.

### DIFF
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -72,3 +72,34 @@
         Archived {{ item.archived | length }} files
         from {{ ansible_hostname }}:{{ item.expanded_paths | join(',') }}
     with_items: "{{_decommission_archive_management.results}}"
+
+
+- hosts: "{{ idr_environment | default('idr') }}-searchengine-hosts"
+  vars:
+    logs:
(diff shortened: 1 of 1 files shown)


sbesson (approved):
Looks good at a glance. Thanks for extending this decommission playbook Khaled. We'll test it during the decommission of `prod114`.

sbesson (changes requested):
While testing this playbook on `prod114`, I needed the following change to avoid an undefined variable

```diff
diff --git a/ansible/decommission/archive-logs.yml b/ansible/decommission/archive-logs.yml
index 9cc7193..b301172 100644
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -3,7 +3,7 @@
 - hosts: >
     {{ idr_environment | default('idr') }}-proxy-hosts
     {{ idr_environment | default('idr') }}-management-hosts
-
+    {{ idr_environment | default('idr') }}-searchengine-hosts
   tasks:
   - name: Get archive prefix
     set_fact:
```

Can you add this change to the PR?

khaledk2:
I have added the change and pushed it.

PR #399, opened 2023-03-23
Add openstack_idr_keypairs role to pilot provisioning playbook

sbesson:
This should allow new key pairs to be uploaded/used while creating new pilots


dominikl (approved):
👍 Tested while spinning up pilot-idr0148.

PR #400, opened 2023-05-26
bump idr-gallery 3.10.2

will-moore:
See https://github.com/IDR/idr-gallery/blob/master/CHANGELOG.md#3100-may-2023
cc @jburel

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -239,7 +239,7 @@ omero_web_config_set:
 omero_web_apps_packages:
 - omero-mapr==0.5.1
 - omero-iviewer==0.11.3
-- idr-gallery==3.9.2
+- idr-gallery==3.10.0
 - omero-figure==4.4.3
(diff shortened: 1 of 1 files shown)


sbesson:
I assume the target here is the next production release of IDR? Worth deploying on `idr-testing` first?

will-moore:
I'm currently deploying https://github.com/IDR/idr-gallery/pull/34 on `idr-testing`, but can easily restore that once we've tested the release there 👍

sbesson (approved):
Happy for this to be deployed on `idr-next` if this has been tested previously and we want to perform a final check there. Who would be best suited for the functional review?

will-moore:
I think @francesw and @pwalczysko have tested the PR (https://github.com/IDR/idr-gallery/pull/31) but I'm also happy to test and check the deployment looks good.

sbesson:
This is now deployed on `prod116`. Assigning people for a final functional assessment of the fix.

PR #408, opened 2023-08-15
Disable TLS 1.0 & 1.1 and use ffdhe2048 DHE group

sbesson:
See https://github.com/ome/omero-certificates/pull/36 for the wider context, this commit manually ports the minimal set of changes for CentOS 7 (disabling TLS 1.0/1.1 and using  the pre-defined DH groups ffdhe2048.

These changes should be reviewed once a new version of `omero-certificates` is released with these changes and TLS 1.3 enabling built-in.


sbesson:
Due to the complexity, closing and deploying manually for testing purposes

PR #409, opened 2023-09-04
Add post OMERO deployment step to install omero-pc-steward

sbesson:
This is only a requirement for the readwrite OMERO server where imports will be performed


sbesson:
Deployed on `prod119`

PR #410, opened 2023-10-05
Update EBI NFS share playbook to also mount S3 buckets

sbesson:
- Install [goofys](https://github.com/kahing/goofys) under /usr/local/bin
- Install fuse
- Use Ansible mount command to mount the bucket using `/etc/fstab`
- Take an input dictionary defining the bucket, mount points and options


dominikl (approved):
Looks good. Seems to work on the pilot-idr0139 👍

PR #414, opened 2024-02-02
Remove deployment of EFK stack on IDR systems

sbesson:
Unlike the Prometheus/Grafana stack, this experimental stack seems to be largely unused at the moment. Additionally, the versions of the different components are out-of-date. Rather than looking into their upgrade with the migration to Rocky Linux 9, this proposes to remove the EFK stack from the monitoring playbook.


sbesson:
Deployed on `test120`

joshmoore:
A heads up that with the addition of microservices on idr-testing that there's Yet-Another log file location that needs checking to find possible stacktraces. It might be that this infrastructure then becomes more relevant.

sbesson:
Discussed briefly at the IDR meeting, @khaledk2 will assess whether EFK is useful for the micro-services deployment. Holding off for the `prod120` redeployment

PR #422, opened 2024-05-20
Upgrade to use PostgreSQL 16

khaledk2:
Upgrade to PostgreSQL 16


sbesson (changes requested):
Can you also update https://github.com/IDR/deployment/blob/50983fa53235c336cd8b283edd0b78691c179937/ansible/decommission/archive-instance-services.yml#L21 ?

PR #424, opened 2024-06-03
Set value for effective cache size and use PR #41 for ome.postgresql

khaledk2:
This PR sets a value for the effective_cache_size  attribute (75% of the physical machine memory ) and uses https://github.com/ome/ansible-role-postgresql/pull/41

### DIFF
--- a/ansible/group_vars/database-hosts.yml
+++ b/ansible/group_vars/database-hosts.yml
@@ -38,3 +38,4 @@ postgresql_install_extensions: True
 
 postgresql_server_conf:
   shared_buffers: "{{ (ansible_memtotal_mb / 4) | int }}MB"
+  effective_cache_size: "{{ (ansible_memtotal_mb * 0.75 ) | int }}MB"
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -90,8 +90,9 @@
 - name: ome.openstack_volume_storage
   version: 2.0.0
 
-- src: ome.postgresql
-  version: 5.4.0
+- name: ome.postgresql
+  src: https://github.com/khaledk2/ansible-role-postgresql
+  version: using_the_orginal_postgres_config_file
 
 - name: ome.postgresql_client
   version: 0.4.3


khaledk2:
I think the tests have failed because of the requests version 2.32.0 bug. I have modified the workflow to install "requests < 2.32.0"

sbesson:
Deployed as `test122b` for testing. Note the configuration file has multiple duplicate configurations

```
[sbesson@test122b-database ~]$ sudo tail -n 15 /var/lib/pgsql/16/data/postgresql.conf
#------------------------------------------------------------------------------
# CUSTOMIZED OPTIONS
#------------------------------------------------------------------------------

# Add settings for extensions here

listen_addresses = '*'

shared_buffers = 8002MB


listen_addresses = '*'

effective_cache_size = 23880MB
shared_buffers = 7960MB
```

khaledk2:
I have updated the Postgres Ansible role  (https://github.com/ome/ansible-role-postgresql/pull/41#issuecomment-2195491374), so it will account for this situation and the added variables will not be duplicated by the end of the configuration file.

sbesson:
Proposing to deploy this as part of `test123` /cc @jburel

khaledk2:
The effective_cache_size value on `prod122` is 24GB

```
sudo cat /var/lib/pgsql/16/data/postgresql.conf | grep effective_cache_size
effective_cache_size = 24GB
```
We have changed this value manullay on `prod121 `after upgrading to `PostgreSQL 16`.
https://github.com/ome/ansible-role-postgresql/pull/41#issuecomment-2144877965

I could not find the container which was used to index the data on the `idr-prod122`.

So, to have the required comparison, I have run the indexer on the `idr-testing123` with the `effective_cache_size=23880MB`. The indexing time is:
`7hrs and 31 minutes`

I have set the `effective_cache_size`  value to the default one, i.e. `4GB `and restarted the database server then run the indexer again. The indexing time is:
``12hrs and 51 minutes
``

sbesson (commented):
Thanks for the report. Assuming there are no other performance measurements expected by the IDR, I think the next steps would be for to review, merge & release the upstream role and update this PR.

As a possible follow-up, the IDR PostgreSQL configuration file could be cleaned up for better readability / maintenance by commenting the configuration lines which are identical to the default or overridden at the bottom of the file like `shared_buffers` and `effective_cache_size`.

PR #425, opened 2024-06-10
Fix requests 2.32.0 issue

khaledk2:
This PR capps the version of requests to fix the requests 2.32.0 issue. 
cc @will-moore


sbesson (approved):
Thanks @khaledk2 this is a blocker to getting new changes in so let's get this merged as soon as the builds are green

PR #428, opened 2024-07-01
PostgreSQL: set effective_cache_size and increase max_connections

sbesson:
The first configuration change was found to be the critical parameter modified during the Rocky Linux 9 migration. The maximal number of connections was bumped during the work on test122

For the next release, we might want to resurrect the work on #424 but the underlying Ansible changes requires additional assessment. In the scope of the `prod122` release, this should contain the minimal set of PostgreSQL variable changes to allow deployments without manually copying a configuration file over.


will-moore (approved):
Looks good.
cache size matches change in https://github.com/IDR/deployment/pull/424/files and the max_connections corresponds to changes as discussed by IDR team. 👍

sbesson:
Closing in favor of #424 which will is expected to be deployed on the next production

PR #429, opened 2024-07-30
Bump ZarrReader to 0.5.2 and omero-server to 5.6.12

dominikl:
See title.

### DIFF
--- a/ansible/group_vars/omero-hosts.yml
+++ b/ansible/group_vars/omero-hosts.yml
@@ -9,7 +9,7 @@
 idr_omero_server_release: 5.6.12
 omero_server_checkupgrade_comparator: '!='
 
-idr_omero_web_release: 5.27.0
+idr_omero_web_release: 5.26.0
 # omero-web depends on omero-py but may not pin the latest release
 # omero_web_python_addons:
 #   - omero-py==5.9.0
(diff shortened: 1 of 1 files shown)


dgault (approved):

sbesson (approved):
Deployment changes look good to me. From this morning's IDR meeting, I understand the goal is to have this deployed onto idr-testing alongside the image-region micro-service with the same version of ZarrReader for a new round of testing.

At the code level, this raises the question of whether the micro-service addition will come as a separate PR and be included here. Note that from an deployment perspective, it should be also possible to update `prod122` with these changes first and schedule the integration of the micro-service at a later date.

dgault (changes requested):

sbesson:
The following playbooks `idr-omero.yml`, `idr-omero-readonly.yml` and `idr-omero-web.yml` have been successfully executed against `prod122` upgrading the server and web environments as required.

Note that as part of this action, the PostgreSQL configuration file was "downgraded" to the state installed by the `ome.postgresql:5.4.0` role. @khaledk2 manually added `effective_cache_size = 24GB` to the configuration file which was the primarily change from the CentOS 7 -> Rocky Linux 9 upgraded. Mid-term, this should be addressed properly by completing the work on #424.

PR #431, opened 2024-08-16
Bump OMERO.server to version 5.6.13

sbesson:
See https://www.openmicroscopy.org/2024/08/15/omero-5-6-13.html

No functional change, this is primarily a tidy-up


sbesson:
Deployed on `prod123`

PR #434, opened 2024-09-30
Add antibody name to mapr search

dominikl:
See title. For testing this probably has to be applied to pilot-idr0159, because the field "Antibody Name" doesn't exist in other deployments (renamed from "Antibody").
/cc @sbesson


sbesson:
That has been now deployed on `pilot-idr0159`

dominikl:
Thanks @sbesson . Looks good to me. What do you think @francesw ?

sbesson (approved):

francesw:
:+1: looks good

sbesson:
Merging for inclusion in `prod124`

PR #437, opened 2024-10-08
Comment out idr0158 redirect

dominikl:
See title. I just commented it out, might be handy to have it for other/future redirects. 
/cc @sbesson 

Bit worried to apply that on idr-next... would it overwrite other configs (e.g. nginx) again!?


sbesson (approved):
Should be fine I'll apply it and make sure it does not update OMERO VMs

sbesson:
Deployed on `prod124`

dominikl:
👍  Thanks Seb!

Issue #436, opened 2024-10-07
Micro-service deployment & NGINX configuration


sbesson:
The deployment of the image-region micro-service was added in https://github.com/IDR/deployment/pull/430 via the extension the `idr-omero-readonly.yml` playbook.

Because the role is internally patching the NGINX configuration created by the `ome.omero_web` role, a possible issue is that executing the `idr-omero-web.yml` playbook in isolation  can revert this change. This is effectively what happened in https://github.com/IDR/deployment/pull/435 in order to deploy an OMERO.map configuration change

There are a few possible options to make the systems more robust:
- either consider the sub-playbooks should no longer be executed individually and only `idr-01-install-idr.yml` should be executed when modifying a deployment. An advantage of this approach is that it should ensure that all playbooks are run in the order. A downside is the execution time. In particular, tackling https://github.com/IDR/deployment/issues/433 might be useful if this is the approach
- move the `ome.omero_ms_image_region` role to the `idr-omero-web.yml` playbook so that it can always be executed after the `ome.omero_web` role and prevent any regression in the NGINX configuration

/cc @dominikl

sbesson:
See https://github.com/IDR/deployment/pull/438#issuecomment-3057361753 for a third option which is to entirely disabled the role level NGINX configuration and have this fully managed at the level of the IDR playbook

PR #438, opened 2024-10-14
Update Nginx Configuration to use Microservice  after updating OMERO-Web

khaledk2:
This PR fixes the issue https://github.com/IDR/deployment/issues/436 .
I have added a check on the `idr-omero-web` playbook for Micros-services user existence, i.e., `omero-server.` If so, it runs the `omero_ms_image_region` role to update the Nginx configuration file for MS.
The MS deployment cannot be moved to this playbook as the MS user is created after running this playbook.

The molecule test failed without modification due to the failure of starting `iptables `service
https://github.com/khaledk2/deployment/actions/runs/11316428331/job/31468697284#step:5:1497
I have checked this issue and found out that it failed due to this error:
`/sbin/iptables does not exist`
I have checked the `iptables_raw `role and successfully ran the molecule test. I compared the installation packages on the role molecule test and the installation package on the deployment playbook, and they are not identical; more packages are installed on the role which are related to `iptables-nft.` 
I have changed the installation package on the iptables_raw from `iptables-services` to `iptables-nft-services` and updated the deployment requirements to point to this modified branch (https://github.com/khaledk2/ansible-role-iptables-raw/tree/refs/heads/upgrade_package_name).
Then I ran the molecule tests in the deployment repo and it passed.
I will open PR for the `iptables_raw` role  shortly

### DIFF
--- a/ansible/idr-omero-web.yml
+++ b/ansible/idr-omero-web.yml
@@ -10,3 +10,23 @@
   # Vars are in group_vars/omero-hosts.yml
 
   environment: "{{ idr_ANSIBLE_ENVIRONMENT_VARIABLES | default({}) }}"
+
+# This bloock will run in case of updating omero-web
+# It will check for the ms user exists
+# If so, it will run the role to update the omero web nginx configuration file
+
+- hosts: "{{ idr_environment | default('idr') }}-omeroreadonly-hosts"
+  become: true
+
+  pre_tasks:
+  - name: check if ms user exist
+    ansible.builtin.user:
+      name: "omero-server"
+    check_mode: true
+    register: service_user
+
+  roles:
+  - role: ome.omero_ms_image_region
+    when: "{{ service_user.state | d('') == 'present' }}"
+
+  environment: "{{ idr_ANSIBLE_ENVIRONMENT_VARIABLES | default({}) }}"
--- a/ansible/requirements.yml
+++ b/ansible/requirements.yml
@@ -39,8 +39,9 @@
 - name: ome.ice
   version: 4.4.4
 
-- src: ome.iptables_raw
-  version: 0.4.0
+- name: ome.iptables_raw
+  src: https://github.com/khaledk2/ansible-role-iptables-raw
+  version: upgrade_package_name
 
 - src: ome.java
   version: 2.2.0


sbesson (commented):
Thanks for looking into these issues, there are at least three separate issues here:

- the `ome.iptables_raw` service startup error I can reproduce in https://github.com/sbesson/deployment/actions/runs/11435648689/job/31811454141. I have left a few comments on https://github.com/ome/ansible-role-iptables-raw/pull/13#pullrequestreview-2381242848 about the underlying change. We might need a full testing environment to check the firewall rules are working as expected
- the `ome.omero_web/ome.omero_ms_image_region` dependency. As shown in #436 although the two roles are built separately, there is a strong dependency between the two as the second one patches the NGINX configuration file created by the first one. In practice, they should always be run at the same time with the current architecture design. For future additions, this should probably be rethought to either separate the NGINX configuration from the application roles or allow to specify NGINX extension points.
- the `ome.omero_ms_image_region/ome.omero_server` dependency. The default user of the former role is the `omero-server` user created by the latter. Although the proposed check introduced in this PR might work, I think we need to find a different solution as this is adding another layer of complex logic to workaround the fact we have a circular dependency. 

I think the order in https://github.com/IDR/deployment/blob/87c62108b953cbe3d30dc20ac984bb958eac15fe/ansible/idr-01-install-idr.yml#L9-L12 is currently incorrect. Based on the dependencies expressed above, my understanding is that this would be the required steps:
1- `idr-omero.yml` i.e. install OMERO.server on the read-write servr
2- `idr-omero-readonly.yml` i.e. install OMERO.server on the read-only servers
3- `idr-omero-web.yml` i.e. install OMERO.web on all servers and the image-region micro-service on the read-only services

dominikl:
I get past the iptables issue, but now stuck at: 
```
PLAY [Deploying search engine] ******************************************************************************************************************************************************************************

TASK [Gathering Facts] **************************************************************************************************************************************************************************************
fatal: [19cd7a52-3664-4ed9-8c5f-f0b880e28221]: UNREACHABLE! => {"changed": false, "msg": "EOF on stream; last 100 lines received:\n", "unreachable": true}

PLAY RECAP **************************************************************************************************************************************************************************************************
19cd7a52-3664-4ed9-8c5f-f0b880e28221 : ok=0    changed=0    unreachable=1    failed=0    skipped=0    rescued=0    ignored=0
```

sbesson:
From yesterday's meeting on the upcoming changes for `prod125`, we discussed upcoming updates to the deployment playbooks to include the latest gallery application as well as configuration changes. /cc @will-moore 

Given this work requires additional changes/testing, could the `iptables` fix be extracted into a standalone PR so that we can get it tested separately and fix the deployment?

sbesson:
`iptables` has been fixed in https://github.com/IDR/deployment/pull/440.

Back to the source of this issue, I am going to suggest a different approach here. The root of the problem is that the `ome.omero-web` role internally use of `omero web config nginx` to configure NGINX, effectively assuming  standard OMERO.web deployment and does not include the flexibility to customise this NGINX configuration.

This is not the case when micro-services or other customisations are introduced as the generated NGINX configuration needs to be amended with new app servers and new endpoints. At the moment, we are working around this approach by patching the NGINX configuration in `ome.omero_ms_image_region` but this approach is fragile at best and also makes these sub playbook no longer idempotent.

Instead, I think the best practice given the current role API is to disable the automatic NGINX configuration via `omero_web_setup_nginx: False` as well as its patching via `omero_ms_image_region_update_nginx: False` and specify and deploy the NGINX configuration entirely at the level of the IDR deployment playbooks.

PR #446, opened 2025-02-17
update searchengine version

khaledk2:
This pull request updates the searchengine to version 0.6.0.


sbesson (approved):

sbesson:
As discussed at the IDR meeting, this is fine to merge and include in follow-up deployments. For `prod126` which has already been deployed, past experience with updating the searchengine has caused some corruption in the ElasticSearch that had to be resolved manually. So happy for the latest container to be updated manually on the `prod126-searchengine instance

PR #448, opened 2025-05-12
Modify Deployment for multi data source searchengine

khaledk2:
This PR  contains the required changes to deploy the multi-data source searchengine https://github.com/ome/omero_search_engine/pull/102

### DIFF
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -10,7 +10,6 @@
     set_fact:
       decommission_archive_prefix: "{{ ansible_hostname }}-{{ ansible_date_time.date | replace('-', '') }}"
 
-
 - hosts: "{{ idr_environment | default('idr') }}-proxy-hosts"
   vars:
     logs:
@@ -41,7 +40,6 @@
         Archived {{ item.archived | length }} files
         from {{ ansible_hostname }}:{{ item.expanded_paths | join(',') }}
     with_items: "{{_decommission_archive_proxy.results}}"
-
 
 - hosts: "{{ idr_environment | default('idr') }}-management-hosts"
   vars:
@@ -74,11 +72,11 @@
         from {{ ansible_hostname }}:{{ item.expanded_paths | join(',') }}
     with_items: "{{_decommission_archive_management.results}}"
 
-
 - hosts: "{{ idr_environment | default('idr') }}-searchengine-hosts"
   vars:
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
+      - {"name": "searchengine_backup_data/", path: "/searchengine_backup/backup_data"}
   tasks:
 
   - name: Archive searchengine logs
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -77,7 +77,7 @@
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
       # We may move archive the searchengine indexed data snapshot to a separate playbook
-      - {"name": "searchengine_backup_data/", path: "/searchengine_backup"}
+      - {"name": "searchengine_backup_data", path: "/searchengine_backup"}
   tasks:
 
   - name: Archive searchengine logs
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -76,7 +76,6 @@
   vars:
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
-      # We may move archive the searchengine indexed data snapshot to a separate playbook
       - {"name": "searchengine_backup_data", path: "/searchengine_backup"}
   tasks:
 
--- a/ansible/decommission/archive-logs.yml
+++ b/ansible/decommission/archive-logs.yml
@@ -76,7 +76,8 @@
   vars:
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
-      - {"name": "searchengine_backup_data/", path: "/searchengine_backup/backup_data"}
+      # We may move archive the searchengine indexed data snapshot to a separate playbook
+      - {"name": "searchengine_backup_data/", path: "/searchengine_backup"}
   tasks:
 
   - name: Archive searchengine logs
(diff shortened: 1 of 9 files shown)


ansible/decommission/archive-logs.yml
```diff
@@ -74,11 +72,12 @@
         from {{ ansible_hostname }}:{{ item.expanded_paths | join(',') }}
     with_items: "{{_decommission_archive_management.results}}"
 
-
 - hosts: "{{ idr_environment | default('idr') }}-searchengine-hosts"
   vars:
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
+      # We may move archive the searchengine indexed data snapshot to a separate playbook
+      - {"name": "searchengine_backup_data/", path: "/searchengine_backup"}
```
sbesson, line 80: Should it be `"searchengine_backup_data"` ?

ansible/idr-searchengine.yml
```diff
@@ -151,6 +161,35 @@
...
+      cleanup: True
+      #auto_remove: yes
+      command: "set_automatic_refresh -a {{ automatic_refresh }}"
+      state: started
+      volumes:
+        - "{{ apps_folder }}/searchengine/searchengine/:/etc/searchengine/"
+
+# restore the elastic indexed data from the backup
+  # This required that the  indexed data snapshot be copied to the
+  # backup folder before running the  elasticsearch deployment playbook
```
sbesson, line 180: For the OMERO database, binary repository and NGINX cache, the volumes are being cloned from the former production deployment and used during the provisioning step.
Is it possible to follow the same strategy here?

khaledk2: We need to transfer the snapshot backup data from the current production server to the next production server and restore it on the new machine. Would it be possible to mount a small 20 GB volume for the backup data, which we could then clone for deployment?

sbesson: A dedicated volume would certainly be consistent with the pattern established for cloning the database and the OMERO binary repository from the previous production environment to a new production or a pilot.

https://github.com/IDR/deployment/blob/f7e8bb9699a64cb291c8dde09f519bb7c0eaf756/ansible/openstack-create-publicidr.yml#L151-L171 will need to be adjusted to declare a new volume and https://github.com/IDR/deployment/blob/master/scripts/os-idr-snapshot.sh will also need to be updated to snapshot this new volume.

ansible/decommission/archive-logs.yml
```diff
@@ -74,11 +72,12 @@
         from {{ ansible_hostname }}:{{ item.expanded_paths | join(',') }}
     with_items: "{{_decommission_archive_management.results}}"
 
-
 - hosts: "{{ idr_environment | default('idr') }}-searchengine-hosts"
   vars:
     logs:
       - { name: 'searchengine', path: '/data/searchengine/searchengine/logs' }
+      # We may move archive the searchengine indexed data snapshot to a separate playbook
```
sbesson, line 79: I don't understand this comment.
Is this a something that should be captured as an issue for the future? Under what conditions should we or should we not archive the backup on decommissioning?

ansible/group_vars/searchengine-hosts.yml
```diff
@@ -2,19 +2,22 @@ apps_folder: /data
...
 database_name: idr
+default_datasource: idr
 database_username: omeroreadonly
 database_user_password: "{{ idr_secret_postgresql_password_ro | default('omero') }}"
 searchenginecache_folder: /data/searchengine/searchengine/cacheddata/
 search_engineelasticsearch_docker_image: docker.elastic.co/elasticsearch/elasticsearch:8.8.1
-searchengine_docker_image: openmicroscopy/omero-searchengine:0.6.0
-#ansible_python_interpreter: path/to/bin/python
+searchengine_docker_image: openmicroscopy/omero-searchengine:0.7
+#ansible_python_interpreter: path/to/bin/python`
```
sbesson, line 11: Type

ansible/idr-searchengine.yml
```diff
@@ -167,4 +206,4 @@
       restart_policy: always
       volumes:
         - "{{ apps_folder }}/searchengine/searchengine/:/etc/searchengine/"
-        - "{{ apps_folder }}/searchengine/searchengine/cachedata:/etc/searchengine/cachedata"
+        - "{{ apps_folder }}/searchengine/searchengine/cachedata:/etc/searchengine/cachedata"
```
sbesson, line 209: Missing newline

sbesson (changes requested):
A few comments. Also two general process questions:

- what happens if the `searchengine_backup` volume is created but not cloned from a previous volume? Will the playbook still initialise an empty cluster or will it fail during `restore_elasticsearch_data` command? This will be immediately relevant for the creation of `prod128` as the volume will not be cloned from a previous version.
- what is the process for creating a new backup of the search cluster? Does that happen automatically after the indexer is executed? Or is that a separate command?

We will probably need to 2 companion PRs:
- one against submission workflow with the indexing/backup commands if they are different from the current one
- one against the deployment workflow to update the creation of new volumes and include the new volume to clone

sbesson:
> This one should run after the deployment playbooks have completed successfully. It will check for the existence of the snapshot before running

Shouldn't it run automatically as part of the search engine deployment then? Is the task asynchronous? If not how long will it typically take?

> This playbook should run just before releasing the production server.

So the current indexing process is unchanged to update for new studies. Should this be executed by the person running the indexer using the same approach `docker run` rather than requiring an Ansible set-up?

khaledk2:
The search engine data restore process is asynchronous, and it runs automatically after the last commit.
Indexing and cache updates for newly added studies are performed manually.
Backups can also be run manually using the following command

`docker run  -v /data/searchengine/searchengine/:/etc/searchengine/  -v /data/searchengine/searchengine/logs/:/opt/app-root/src/logs/ --network searchengine-net openmicroscopy/omero-searchengine:0.7 backup_elasticsearch_data`

sbesson:
> Backups can also be run manually using the following command
>
>> docker run  -v /data/searchengine/searchengine/:/etc/searchengine/  -v /data/searchengine/searchengine/logs/:/opt/app-root/src/logs/ --network searchengine-net openmicroscopy/omero-searchengine:0.7 backup_elasticsearch_data

@jburel @francesw @will-moore @dominikl what are your thoughts about the best place to integrate this into the IDR lifecycle?

sbesson (commented):
Let's discuss the state of this PR tomorrow at the IDR weekly meeting. I think we need to clarify two questions before spinning up `prod128` with this included:

- the process for creating a dump of the search engine as raised in https://github.com/IDR/deployment/pull/448#issuecomment-2959250030. My opinion is that this task is outside the scope of these playbooks
- re-reading the diff, what would happen if the playbooks and notably `idr-02-services.yml` was re-run against an environment previously deployed? Would the restore command be re-executed and what would be the expected state of the indexer?

PR #449, opened 2025-05-26
Only redirect specific /study/idr0083/figure urls. Ignore other /stud…

will-moore:
This aims to address issues found at https://github.com/IDR/idr-gallery/pull/50 where we want to create new landing pages for every study at e.g. `idr.openmicroscopy.org/study/idr0001/`.

This currently fails because all `/study/` URLs are redirected because we want to support https://idr.openmicroscopy.org/study/idr0083/figure/4i and https://idr.openmicroscopy.org/study/idr0083/figure/4r redirects.

This PR attempts to be more specific, ONLY handling those 2 urls and not redirecting other `/study` urls.

I don't have an easy way to test this. This was created with the help of ChatGPT!

cc @sbesson

### DIFF
--- a/ansible/group_vars/proxy-hosts.yml
+++ b/ansible/group_vars/proxy-hosts.yml
@@ -239,8 +239,13 @@ nginx_proxy_redirect_map:
   dest: /webclient/img_detail/9822152/?dataset=10201&x=80560&y=77440&zm=66&c=1|495:9204$808080&m=g
 
 nginx_proxy_redirect_map_locations:
+# These are the only /study/ URLs that should be redirected
+- location: "= /study/idr0083/figure/4i"
+  code: 301
+- location: "= /study/idr0083/figure/4r"
+  code: 301
 # TODO: change to 301 when we're happy
-- location: "~ ^/(mito|tara|pgpc|study)($|/)"
+- location: "~ ^/(mito|tara|pgpc)($|/)"
   code: 302
 
 # "= /" has higher priority than "/" in the proxy config


sbesson:
Which environment should this be deployed against for functional testing?

will-moore:
I think any environment could be used to confirm that `/study/` urls don't redirect and `/study/idr0083/figure/4i` behave as before, but idr-testing also has the landing pages PR deployed, so that would be handy if possible to test both together.

will-moore:
NB: those URLs (e.g. `/study/a`) are redirecting to `/webclient/?experimenter=-1` on idr-testing instead of the home page (which was the previous behaviour).

will-moore:
Also, if I do `$ ssh -A idr-testing.openmicroscopy.org -L 1080:omeroreadwrite:80` then landing pages are working at e.g. http://localhost:1080/study/idr0003/ - don't know if that helps to understand what's up here?

sbesson:
Is https://github.com/IDR/idr-gallery/pull/50 also deployed on `test125` ? In that case, I would expect as a result of this `/study/*` is no longer redirected at the proxy level but instead forwarded to OMERO.web and handled via the `idr-gallery` application.

will-moore:
@sbesson yes, on idr-testing I ran:

```
for server in omeroreadwrite omeroreadonly-1 omeroreadonly-2 omeroreadonly-3 omeroreadonly-4; do ssh $server "sudo /opt/omero/web/venv3/bin/pip uninstall -y idr-gallery && sudo /opt/omero/web/venv3/bin/pip install git+https://github.com/IDR/idr-gallery.git@refs/pull/50/head && sudo service omero-web restart"; done
```

I would also expect `/study/*` not to be redirected, but that's what I'm seeing! 👎

sbesson:
Merging for inclusion in `test128`

PR #452, opened 2025-06-16
initialize and mount the new volume

khaledk2:
Initialize and mount the new `searchengine_backup `volume

### DIFF
--- a/ansible/os-idr-volumes.yml
+++ b/ansible/os-idr-volumes.yml
@@ -33,5 +33,5 @@
      {{ idr_environment | default('idr') }}-searchengine-hosts
   roles:
   - role: ome.storage_volume_initialise
-    storage_volume_initialise_device: "{{ elasticsearch_backup_folder_vol_dev | default('/dev/vdb1') }}"
+    storage_volume_initialise_device: "{{ elasticsearch_backup_folder_vol_dev | default('/dev/vdc') }}"
     storage_volume_initialise_mount: "{{ elasticsearch_backup_folder }}"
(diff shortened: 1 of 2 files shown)


ansible/os-idr-volumes.yml
```diff
@@ -28,3 +28,10 @@
   - role: ome.storage_volume_initialise
     storage_volume_initialise_device: "{{ data_vol_dev | default('/dev/vdb') }}"
     storage_volume_initialise_mount: /data
+
+- hosts: >
+    - hosts: "{{ idr_environment | default('idr') }}-searchengine-hosts"
```
sbesson, line 33: Duplicate -`hosts:`

sbesson (changes requested):

sbesson (changes requested):

sbesson:
```
ERROR! We were unable to read either as JSON nor YAML, these are the errors we got from each:
JSON: Expecting value: line 1 column 1 (char 0)

Syntax Error while loading YAML.
  found unacceptable key (unhashable type: 'AnsibleMapping')

The error appears to be in '/tmp/management_tools/idr/deployment/ansible/os-idr-volumes.yml': line 37, column 39, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    storage_volume_initialise_device: "{{ elasticsearch_backup_folder_vol_dev | default('/dev/vdb1') }}"
    storage_volume_initialise_mount: {{ elasticsearch_backup_folder }}
                                      ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"

```

PR #453, opened 2025-06-19
Change the Elasticsearch backup folder ownership

khaledk2:
This PR changes the ownership of the Elasticsearch backup folder to resolve the access permission issue I encountered while attempting to create the backup.

### DIFF
--- a/ansible/idr-elasticsearch.yml
+++ b/ansible/idr-elasticsearch.yml
@@ -15,6 +15,7 @@
       path: "{{ elasticsearch_backup_folder }}"
       owner: 1000
       group: root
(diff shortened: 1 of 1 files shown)


sbesson:
I am going to close this as per https://github.com/IDR/deployment/pull/453#pullrequestreview-2943418744.

The situation is similar to other volumes like the database one which is owned by `postgres` rather than `root` where we do not run a `chown -R` as part of the deployment. The real test will be the deployment of `prod129` which should inherit the ownership of `prod128-searchengine-backup`. If this causes issues, we can evaluate how to update our deployment infrastructure

PR #458, opened 2025-09-01
Deploy the search engien with Nginx and Gunicorn

khaledk2:
This PR modifies the searchengine deployment to work with Nginx. This is primarily intended to facilitate work on the searchengine, `Dump data and Query results in bff format` PR (https://github.com/ome/omero_search_engine/pull/114#issuecomment-3225686937)


khaledk2:
Nginx has been installed and deployed on the `test128-searchengine`, and the main idr proxy has been updated to work with it.

